PR #1, opened 2017-09-07
Add new card ID

retsamedoc:
New Card ID (0x6811-> Curaco PRO, R7 370)


vandroiy2013:
Sorry. This model repository is ONLY for device-id's from native Apple kexts.

retsamedoc:
If that is the case, I don't understand why WhateverGreen is even needed.

PR #4, opened 2018-07-15
Add framebuffer patches

headkaze:


WhateverGreen/kern_igfx.cpp
```diff
@@ -248,3 +349,493 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
...
+		callbackIGFX->framebufferPatches[i].find->release();
+		callbackIGFX->framebufferPatches[i].replace->release();
+	}
+	
+	return r;
+}
+
+uint8_t *IGFX::findFramebufferId(uint32_t framebufferId, uint8_t *startingAddress, size_t maxSize) {
+	uint8_t *startAddress = startingAddress;
+	uint8_t *endAddress = startingAddress + maxSize;
```
vit9696: This should be `uint8_t *endAddress = startingAddress + maxSize - sizeof(uint32_t);` or you risk addressing invalid memory. While unlikely, it is best to stay safe.

WhateverGreen/kern_igfx.cpp
```diff
@@ -248,3 +349,493 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
...
+			framebufferAddress = callbackIGFX->findFramebufferId(callbackIGFX->framebufferPatches[i].framebufferId, (uint8_t *)callbackIGFX->gPlatformInformationList, framebufferSize);
+			framebufferId = callbackIGFX->framebufferPatches[i].framebufferId;
+		}
+		
+		KernelPatcher::LookupPatch patch;
+		
+		patch.kext = callbackIGFX->currentFramebuffer;
+		patch.find = (const uint8_t *)callbackIGFX->framebufferPatches[i].find->getBytesNoCopy();
+		patch.replace = (const uint8_t *)callbackIGFX->framebufferPatches[i].replace->getBytesNoCopy();
+		patch.size = min(callbackIGFX->framebufferPatches[i].find->getLength(), callbackIGFX->framebufferPatches[i].replace->getLength());
```
vit9696: It is better to assert that these are equal and skip the patch otherwise. No need to workaround user errors, at best we should SYSLOG and continue.

vit9696:
Merged with minor alterations. Thanks :)

PR #5, opened 2018-07-25
Add CNL ICL support

headkaze:


headkaze:
There's no CNL framebuffer support in macOS 10.15.4. The only new addition is AppleIntelICLLPGraphicsFramebuffer.kext. The ig-platform-id's are listed in /System/Library/Extensions/AppleIntelICLLPGraphicsFramebuffer.kext/Contents/Info.plist/IOKitPersonalities/AppleIntelFramebufferController/IOPCIPrimaryMatch. Which are 0xFF058086 0x8A708086 0x8A718086 0x8A518086 0x8A5C8086 0x8A5D8086 0x8A528086 0x8A538086 0x8A5A8086 0x8A5B8086.

headkaze:
There appears to be more hardware supported since I last dumped the platform list for ICL. I'm trying to get a new dump but it doesn't seem to be working anymore.

From memory all I had to do was use debug versions of Lilu and WEG along with the following boot args:

`
-liludbgall -wegdbg lilucpu=12 -igfxdump
`

And then add a spoofed device-id (0x8A700000):

```
<key>PciRoot(0x0)/Pci(0x2,0x0)</key>
<dict>
	<key>device-id</key>
	<data>cIoAAA==</data>
</dict>
```

I checked in /var/log/ and there is no AppleIntelFramebuffer_X_Y dump file. I also got a KP but that's not out of the ordinary.

Before the KP the Lilu log shows that it has loaded the correct FB kext and successfully wrapped getOSInformation.
```
Lilu      mach: @ (DBG) found executable at path: /System/Library/Extensions/AppleIntelICLGraphics.kext/Contents/MacOS/AppleIntelICLGraphics
...
Lilu   patcher: @ (DBG) loaded kinfo com.apple.driver.AppleIntelICLLPGraphicsFramebuffer at 9 index
...
Lilu   patcher: @ (DBG) wrapped __ZN31AppleIntelFramebufferController16getOSInformationEv
...
WhateverGreen       weg: @ (DBG) configRead32 IGPU 0x00001000 at off 0x02, result = 0x3E928086
WhateverGreen     iokit: @ (DBG) getOSData device-id has 8A70 value
WhateverGreen       weg: @ (DBG) configRead32 reported 0x8a708086 instead of 0x3e928086
```
Maybe it's not possible to get a dump this way anymore? Is there anyone with ICL hardware that can get a framebuffer dump and attach it here?

vit9696:
It is possible and we have a dump. The key is igfxdumpdelay argument introduced in 1.3.9. With something like 10 seconds (the argument is i  milliseconds) you get enough time for the file to save even with the KP to follow.

vit9696:
I also partially fixed the structure for ICL (https://github.com/acidanthera/WhateverGreen/commit/b7e1e49a82fdf265960d5c66e8a0131955f7e3d5), but the flags are all very different now. Some flags were removed, some flags were added, so the same flags for ICL and KBL have different numeric values. Feel free to rebuild the table if you want to work on it.

headkaze:
igfxdumpdelay appears to have a default value of 15000. Is the call to PE_parse_boot_argn() zeroing out the delay? I just commented out the PE_parse_boot_argn() and set delay to 20000 and I got the dump. So either the default should be 20000 or the default should be set if PE_parse_boot_argn() returns false.

vit9696:
It does not, just the default value depends on the hardware you run. For some 15s is enough, for some it is not. I increased it to 20s.

Kmbeatz:
Is there support for the 10th Intel i5 10035g1 with 0x8a568086?

PR #6, opened 2018-09-11
add framebuffer-conX-alldata for bulk replacement of connector data

RehabMan:
This patch adds framebuffer-conX-alldata as discussed here:
https://www.tonymacx86.com/threads/guide-intel-framebuffer-patching-using-whatevergreen.256490/page-14#post-1805692

### DIFF
--- a/WhateverGreen.xcodeproj/project.pbxproj
+++ b/WhateverGreen.xcodeproj/project.pbxproj
@@ -158,6 +158,7 @@
 				1C748C281C21952C0024EED2 /* Products */,
 			);
 			sourceTree = "<group>";
+			usesTabs = 1;
 		};
 		1C748C281C21952C0024EED2 /* Products */ = {
 			isa = PBXGroup;
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -595,16 +595,34 @@ bool IGFX::loadPatchesFromDevice(IORegistryEntry *igpu, uint32_t currentFramebuf
 
 			DBGLOG("igfx", "framebuffer-con%lu-enable %d", i, framebufferConnectorPatchEnable);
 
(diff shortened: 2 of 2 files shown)


vit9696:
Hey, this looks fairly nice to me. Could you please update the changelog and fix indentation (we use tabs in the project, and the change reindents with spaces)? Other than that the diff looks good enough.

We are going to have a release tomorrow, should be able to fit that in.

RehabMan:
Can't update the changelog as I don't know what version you're going to release this as, so maybe you can just fill that in as you do further changelog entries.  Note also I didn't update any version markers anywhere.  With my projects, I don't update the changelog until a build is ready to go, not sure what you're doing with this project (ie. whether changelog reflects changes not yet in the build).

Let me know what your custom Xcode settings are for tabs... I just use default Xcode settings.  Then I guess I can cut/paste to fix tabs vs. spaces.

FYI: There are no diffs due to only space vs. tab.

RehabMan:
One thing I thought of while just reviewing the diffs..., is that my code assumes the connector patch flags ((CPF*). are zeroed to begin with.  Hopefully that is valid assumption.

If not will need code to set all to zero somewhere.

RehabMan:
Note: Maybe the project "Text Settings" should be changed, so it doesn't read "Indent Using Spaces"...

vit9696:
I usually set them [here](https://user-images.githubusercontent.com/4348897/45332174-ba87d580-b577-11e8-8dc2-90a1db1cbca3.png), but you can also fix them on project basis: [example](https://stackoverflow.com/questions/14078455/how-to-change-the-default-code-indentation-in-xcode-4-5). Then repaste the changed lines, which will [fix the offsets](https://github.com/acidanthera/WhateverGreen/pull/6/files#diff-5371124ffcbf43a683698be42b87a515R598).

They are zeroed, do not worry.

PR #8, opened 2018-09-13
add framebuffer-cursormem for patching Haswell specific fCursorMemorySize

RehabMan:
While converting some of my projects, noticed that there was not a semantic patch for fCursorMemorySize in Haswell.

This patch adds framebuffer-cursormem for it.

### DIFF
--- a/WhateverGreen.xcodeproj/project.pbxproj
+++ b/WhateverGreen.xcodeproj/project.pbxproj
@@ -578,7 +578,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.2;
+				MODULE_VERSION = 1.2.3;
 				OTHER_CFLAGS = (
 					"-mmmx",
 					"-msse",
@@ -623,7 +623,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.2;
+				MODULE_VERSION = 1.2.3;
 				OTHER_CFLAGS = (
 					"-mmmx",
 					"-msse",
@@ -754,7 +754,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.2;
+				MODULE_VERSION = 1.2.3;
 				OTHER_CFLAGS = (
 					"-mmmx",
 					"-msse",
--- a/WhateverGreen.xcodeproj/project.pbxproj
+++ b/WhateverGreen.xcodeproj/project.pbxproj
@@ -578,7 +578,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.3;
+				MODULE_VERSION = 1.2.4;
 				OTHER_CFLAGS = (
 					"-mmmx",
 					"-msse",
@@ -623,7 +623,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.3;
+				MODULE_VERSION = 1.2.4;
 				OTHER_CFLAGS = (
 					"-mmmx",
 					"-msse",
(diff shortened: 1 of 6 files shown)


vit9696:
Hi,

Firstly, 1.2.3 is not released yet (and is planned somewhere in Autumn, supposedly October or November closer to 10.14.1), so this is to be included in 1.2.3, not 1.2.4.

Secondly, I think the meta side of the patch should be slightly redesigned: move the common part to a generic template function (basically leave the original function), which works on the framebuffers. Then call another framebuffer-specific function from the inside, which is empty by default and specialised for Haswell to update the cursor memory size.

I am fine with the cursor memory being an external variable in IGFX class, but comment it properly I suppose.

RehabMan:
I'm going to need 1.2.3 for 10.14.0.
I thought it was mentioned 1.2.3 would be released soon (didn't check to verify).
Please consider releasing 1.2.3 by Mojave release, so I don't have to upload my own build.

I was just following the SNB example already in the code, but I'll see if I can find a way to share the code more effectively.

You're thinking something like:
```
	/**
	 *  Extended patching called from applyPlatformInformationListPatch
	 *
	 *  @param frame               pointer to Framebuffer data
	 *
	 */	template <typename T>
	void applyPlatformInformationPatchEx(T* frame) { }
...
// Haswell specialization of applyPlatformInformationPatchEx
template <>
void IGFX::applyPlatformInformationPatchEx(FramebufferHSW *frame) {
	// fCursorMemorySize is Haswell specific
	if (framebufferPatchFlags.bits.FPFFramebufferCursorSize) {
		frame->fCursorMemorySize = fPatchCursorMemorySize;
		DBGLOG("igfx", "fCursorMemorySize: 0x%08X", frame->fCursorMemorySize);
	}
}
...
// from applyPlatformInformationListPatch
	applyPlatformInformationPatchEx(frame);
```

??

Will fix the changelog to indicate 1.2.3.

WhateverGreen.xcodeproj/project.pbxproj
```diff
@@ -578,7 +578,7 @@
 				MODULE_NAME = as.vit9696.WhateverGreen;
 				MODULE_START = "$(PRODUCT_NAME)_kern_start";
 				MODULE_STOP = "$(PRODUCT_NAME)_kern_stop";
-				MODULE_VERSION = 1.2.3;
+				MODULE_VERSION = 1.2.4;
```
vit9696: This and below values are still pointing to the wrong version, I am afraid.

WhateverGreen/kern_igfx.cpp
```diff
@@ -782,14 +783,21 @@ bool IGFX::applyPlatformInformationListPatch(uint32_t framebufferId, Framebuffer
...
+	}
+}
+
 template <typename T>
 bool IGFX::applyPlatformInformationListPatch(uint32_t framebufferId, T *platformInformationList) {
 	auto frame = reinterpret_cast<T *>(findFramebufferId(framebufferId, reinterpret_cast<uint8_t *>(platformInformationList), PAGE_SIZE));
 	if (!frame)
 		return false;
 
-	bool r = false;
```
vit9696: I am afraid it is not the variable here being an error, but rather no use of it in the bottom where we have return true. Should probably be restored, relabled like in the other function, and the return condition should be fixed in the bottom.

WhateverGreen/kern_igfx.hpp
```diff
@@ -110,6 +111,7 @@ class IGFX {
 	 *  Framebuffer hard-code patch
 	 */
 	FramebufferCFL framebufferPatch {};
+	uint32_t fPatchCursorMemorySize;	// Haswell only... and therefore not present in FramebufferCFL
```
vit9696: Please restyle the comment like the one above. Do not want to be picky, but it is important to keep the codebase as clean as possible for own good.

WhateverGreen/kern_igfx.hpp
```diff
@@ -412,6 +414,14 @@ class IGFX {
 	template <typename T>
 	bool applyPlatformInformationListPatch(uint32_t framebufferId, T *platformInformationList);
 
+	/**
+	 *  Extended patching called from applyPlatformInformationListPatch
+	 *
+	 *  @param frame               pointer to Framebuffer data
+	 *
+	 */	template <typename T>
```
vit9696: A newline is missed here.

WhateverGreen/kern_igfx.cpp
```diff
@@ -820,9 +830,10 @@ bool IGFX::applyPlatformInformationListPatch(uint32_t framebufferId, T *platform
 		DBGLOG("igfx", "stolenMemorySize: 0x%08X", frame->fStolenMemorySize);
 		DBGLOG("igfx", "framebufferMemorySize: 0x%08X", frame->fFramebufferMemorySize);
 		DBGLOG("igfx", "unifiedMemorySize: 0x%08X", frame->fUnifiedMemorySize);
-		r = true;
```
vit9696: Ugh, I think this was deleted by accident.

RehabMan: Or rather, not added back...

RehabMan:
Should be ok now after last forced push.

FYI: My intention behind my original 'fix' for the static analyzer was to fix only the static analyzer notifications, not to fix logic bugs that may have been in the original code (original code, the function always returned true, except for the case of framebuffer id not found...)

Let me know if you want me to submit a new pull with clean/1-shot commit history.

vit9696:
Thanks, merged =). Fortunately this place can now squash commits even from the interface, so less people have to bother.

RehabMan:
Neat.

PR #9, opened 2018-09-17
SandyBridge framebuffer uses _PlatformInformationList, not _gPlatformInformationList

RehabMan:
I guess I'm the first to test SandyBridge/HD3000 patching.
Because it doesn't work at all due to AppleIntelSNBGraphicsFB using _PlatformInformationList, not _gPlatformInformationList.
Easy fix in this patch.


vit9696:
Pretty much, huh. Fun enough, that nobody reported it.

PR #11, opened 2018-10-15
CFL backlight patch

headkaze:
Issue is calculation for backlight control in CFL requires writing to BXT_BLC_PWM_FREQ1 and BXT_BLC_PWM_DUTY1 registers. But CFL kext does the calculation in 32-bits which will truncate result and cause a 3 minute black screen upon boot.

Will auto enable CFL backlight patch if CPU gen is CFL and model is laptop. Optional boot flags igfxcflbklt=none / igfxcflbklt=force (or enable-cflbklt IGPU property).

Wrap AppleIntelFramebufferController::hwSetPanelPowerConfig to get original initial value of BXT_BLC_PWM_FREQ1 register and store this back to AppleIntelFramebufferController + 0x2b44 (and in variable for later calculation).

Wrap AppleIntelFramebufferController::hwSetBacklight and AppleIntelFramebuffer::setAttributeForConnection to override writing backlight registers. Calculate in 64-bits so result is not truncated. Set values back to BXT_BLC_PWM_FREQ1 and BXT_BLC_PWM_DUTY1 registers.


WhateverGreen/kern_igfx.cpp
```diff
@@ -198,6 +199,18 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
 		KernelPatcher::RouteRequest request("__ZN9IOService20copyExistingServicesEP12OSDictionaryjj", wrapCopyExistingServices, orgCopyExistingServices);
 		patcher.routeMultiple(KernelPatcher::KernelID, &request, 1);
 	}
+	
+	// Enable CFL backlight patch if CPU gen is CFL and model is laptop
+	cflBacklightPatch = (cpuGeneration == CPUInfo::CpuGeneration::CoffeeLake && WIOKit::getComputerModel() == WIOKit::ComputerModel::ComputerLaptop);
+	cflBacklightPatch = cflBacklightPatch || (info->videoBuiltin && info->videoBuiltin->getProperty("enable-cflbklt"));
```
vit9696: For the IORegistry property there is probably no good reason to shorten it way too much. What about renaming `enable-cflbklt` to `enable-cfl-backlight-fix` = `<01>`?

WhateverGreen/kern_igfx.cpp
```diff
@@ -198,6 +199,18 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
...
+	
+	char igfxcflbklt[128];
+	if (PE_parse_boot_argn("igfxcflbklt", igfxcflbklt, sizeof(igfxcflbklt))) {
+		if (strstr(igfxcflbklt, "none"))
```
vit9696: I think this should be `strcmp(…) == 0`, as we should go for an exact match rather than any string, plus I believe any passed value but `force` should probably be considered equivalent to `cflBacklightPatch = false;` for simplicity?

WhateverGreen/kern_igfx.cpp
```diff
@@ -385,6 +415,75 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
...
+	return FunctionCast(wrapHwSetPanelPowerConfig, callbackIGFX->orgHwSetPanelPowerConfig)(that, arg0);
+}
+
+uint64_t IGFX::wrapHwSetBacklight(void *that, uint32_t arg0) {
+	// hwSetBacklight use our backlight calculation here and skip call to original function to avoid flicker
+	//uint64_t r = FunctionCast(wrapHwSetBacklight, callbackIGFX->orgHwSetBacklight)(that, arg0);
+
+	callbackIGFX->backlightLevel = arg0;
+	
+	// uint64_t bxt_blc_pwm_ctrl1 = reinterpret_cast<uint64_t(*)(void*, uint64_t)>(callbackIGFX->orgReadRegister32)(that, BXT_BLC_PWM_CTL1);
```
vit9696: Declaring symbol address with proper function type will make this a bit more readable, similarly to other plugin code. Should be able to write something like:

```
orgReadRegister32 = solveSymbol<decltype(&orgReadRegister32)>(index, 
	"__ZN31AppleIntelFramebufferController14ReadRegister32Em", address, size);
```

And then use without all those casts, which make things a little tough and long.

WhateverGreen/kern_igfx.cpp
```diff
@@ -385,6 +415,75 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
...
+	// Calculate backlight duty in 64-bits so result is not truncated (essentially CFL backlight fix)
+	bxt_blc_pwm_freq1 = callbackIGFX->backlightFrequency;
+	bxt_blc_pwm_duty1 = static_cast<uint64_t>(callbackIGFX->backlightLevel) * bxt_blc_pwm_freq1 / 0xFFFFLL;
+	
+	reinterpret_cast<uint64_t(*)(void*, uint64_t, uint32_t)>(callbackIGFX->orgWriteRegister32)(that, BXT_BLC_PWM_FREQ1, static_cast<uint32_t>(bxt_blc_pwm_freq1));
+	reinterpret_cast<uint64_t(*)(void*, uint64_t, uint32_t)>(callbackIGFX->orgWriteRegister32)(that, BXT_BLC_PWM_DUTY1, static_cast<uint32_t>(bxt_blc_pwm_duty1));
+	
+	// SYSLOG("igfx", "wrapHwSetBacklight(): BXT_BLC_PWM_CTRL1=0x%X BXT_BLC_PWM_FREQ1=0x%X BXT_BLC_PWM_DUTY1=0x%X backlightLevel=0x%X", bxt_blc_pwm_ctrl1, bxt_blc_pwm_freq1, bxt_blc_pwm_duty1, callbackIGFX->backlightLevel);
+	
+	return 0LL;
```
vit9696: Are you sure it is `uint64_t` and not `IOReturn`, I would expect `kIOReturnSuccess` here, but I have not checked the code yet.

WhateverGreen/kern_igfx.cpp
```diff
@@ -385,6 +415,75 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
...
+}
+
+uint64_t IGFX::wrapSetAttributeForConnection(void *that, uint32_t arg0, uint32_t arg1, uint64_t arg2) {
+	// setAttributeForConnection appears to be called from several places and will cause black screen bug in CFL if
```
vit9696: Same thing about `uint64_t`, and please make it more clear, that the real issue is 32-bit integer truncation in Apple code due to multiplying by a value being too large than usual on CFL. I do not remember the exact values, but was it `0xFFFFF`? Probably worth including sample values here too.

WhateverGreen/kern_igfx.cpp
```diff
@@ -385,6 +415,75 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
 	return FunctionCast(wrapAcceleratorStart, callbackIGFX->orgAcceleratorStart)(that, provider);
 }
 
+size_t IGFX::wrapHwSetPanelPowerConfig(void *that, uint32_t arg0) {
+	// hwSetPanelPowerConfig doesn't have anything to do with the backlight control but it's a
+	// convenient place (low frequency call, and called in AppleIntelFramebufferController::start
+	// before backlight adjustments) to grab the initial c8254 value and patch it into the controller
```
vit9696: Note, that c8254 is BXT_BLC_PWM_FREQ1.

WhateverGreen/kern_igfx.hpp
```diff
@@ -342,7 +429,7 @@ class IGFX {
 	 *  IGHardwareGuC::loadGuCBinary wrapper to feed updated (compatible GuC)
 	 */
 	static bool wrapLoadGuCBinary(void *that, bool flag);
-
+
```
vit9696: Do not want to be picky, but I get a feeling this is some editor artefact.
Also, I believe 2 spaces are used between * and doc description instead of TABs, e.g.:
`*  Original AppleIntelFramebufferController::ReadRegister32 function`.

WhateverGreen/kern_igfx.cpp
vit9696: Was this tested on 10.13.6 special build? If not, perhaps it is best to add a `getKernelVersion() >= KernelVersion::Mojave` check just in case?

vit9696:
The revised versions looks much better, but to be honest I think it is reasonable to wait for the tests and experiments on PNLF changes @RehabMan currently works on here: https://github.com/RehabMan/OS-X-Clover-Laptop-Config/commit/728ade7dca0252e6e311479d594cef779bb49dc9

I am not against merging this (even if with minor alterations), but the maintenance required to update hardcoded FB offset (0x2b44) still strongly upsets me, and I will be interested if a more reliable solution could be crafted by changing the backlight ranges to fit 32-bit unsigned.

Also, I haven't yet looked into the details, but it may make sense to conditionally include the code of this patch (https://bitbucket.org/RehabMan/applebacklightfixup/src/18e6a1ab096fa93119b395c666f8740f232db6b9/AppleBacklightFixup/kern_applbkl.cpp?at=master&fileviewer=file-view-default#kern_applbkl.cpp-38) later. There is not much reason to create many kexts and then try to maintain them from different areas.

vit9696:
Based on the successful research we did the day before, I believe the code should contain the following comment for future generations and my terrible memory, as well as implement boot argument support and avoidance of the hardcoded constant via binary patches.

```C
// Intel backlight is modeled via pulse-width modulation (PWM). See page 144 of:
// https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-kbl-vol12-display.pdf
// Singal-wise it looks as a cycle of signal levels on the timeline:
// 22111100221111002211110022111100 (4 cycles)
// 0 - no signal, 1 - no value (no pulse), 2 - pulse (light on)
// - Physical Cycle (0+1+2) defines maximum backlight frequency, limited by HW precision.
// - Base Cycle (1+2) defines [1/PWM Base Frequency], limited by physical cycle, see BXT_BLC_PWM_FREQ1.
// - Duty Cycle (2) defines [1/PWM Increment] - backlight level,
//   [PWM Frequency Divider] - backlight max, see BXT_BLC_PWM_DUTY1.
// - Duty Cycle position (first vs last) is [PWM Polarity]
// 
// Duty cycle = PWM Base Frequeny * (1 / PWM Increment) / PWM Frequency Divider
// 
// On macOS there are extra limitations:
// - All values and operations are u32 (32-bit unsigned)
// - [1/PWM Increment] has 0 to 0xFFFF range
// - [PWM Frequency Divider] is fixed to be 0xFFFF
// - [PWM Base Frequency] is capped by 0xFFFF (to avoid u32 wraparound), and is hardcoded
//   either in Framebuffer data (pre-CFL) or in the code (CFL: 7777 or 22222).
// 
// On CFL the following patches have to be applied:
// - Hardcoded [PWM Base Frequency] should be patched or set after the hardcoded value is written by patching
//   hardcoded frequencies. 65535 is used by default.
// - If [PWM Base Frequency] is > 65535, to avoid a wraparound code calculating BXT_BLC_PWM_DUTY1
//   should be replaced to use 64-bit arithmetics.
// [PWM Base Frequency] can be specified via igfxbklt=1 boot-arg or backlight-base-frequency property.
```

headkaze:
I've added Faxul's and RehabMan's patches along with new boot args.

- `igfxcflbklt=none` to disable CFL backlight patch
- `igfxcflbklt=opcode` to enable Fraxul's CFL backlight patch
- `igfxcflbklt=wrap` (and `enable-cfl-backlight-fix` property) to enable headkaze's CFL backlight patch
- `igfxcflbklt=freq` to enable RehabMan's CFL backlight patch

Set version to 1.2.4b1 for testing purposes.

We don't have enough testing done on RehabMan's "freq" patch yet. So this commit should allow people to switch between patches easily.

vit9696:
It is not like we are going to fix this for everyone, but @headkaze also had an issue with wake yesterday. I added a workaround to master. Hope it helps.

black-dragon74:
Yeah that too but it affects all 8300H and 8750H processors. So fixing this would mean fixing this for everyone 😊

Going to test this and will report back here.

black-dragon74:
Update:
Scenario 1:
When screen doesn't turn off at initial boot, display works fine. Screen turns back on after sleep.

Senario 2:
If the screen turns off during initial boot you have to put your laptop to sleep and then wake it up again in order to get the backlight.

To sum it all up, backlight issue after sleep cycle that was introduced in 1781c6f0352700a22f49f8cf34ecff2b214ee29f has been fixed. However, the original issue still persists?

We get screen black out 1/4th of the times.

Regards

black-dragon74:
Thanks for these tips @vit9696. Will let you know of the progress.

Also wanted to say this but never could, keep up the good work man! You are a boon to this small hackintosh community. You are an inspiration for young developers like me.

Regards

vit9696:
Um, regarding Camellia, you could have asked actually =)
It is a feature responsible for internal display power management, probably some profile.
You can parse the framebuffers from dumped Intel Framebuffer kext with 010 Editor script (https://github.com/acidanthera/WhateverGreen/blob/master/Manual/IntelFramebuffer.bt).

Another way would be to check our page on applelife (first post, just scroll to CFL graphics):
https://applelife.ru/threads/intel-hd-graphics-3000-4000-4400-4600-5000-5500-5600-520-530-630.1289648/
Look for CameliaDisabled.

black-dragon74:
Hey @vit9696,

I tried with two framebuffers that have Camellia disabled **0x3E9B0000** and **3EA50000** but the issue is still there. Anything else that I should try?

Meanwhile, have you heard from @headkaze regarding the same?

P.S: Using framebuffer with no Camellia meant using an ig-platform-id that has Camellia disabled, right?

Regards

vit9696:
Unfortunate, but I cannot think of anything else. Time will show if this can be i proved, perhaps once headkaze returns.

black-dragon74:
@vit9696 You should have a look at: https://www.tonymacx86.com/threads/bug-black-screen-3-minutes-after-booting-coffeelake-uhd-630.261131/page-33#post-1870840

Regards

vit9696:
Well, it is not like this is something new. In fact, we no longer patch the mathematical operations, so I believe routing `writeRegister32` is not really a necessity for the latest changeset. Will be nice if Fraxul checked whether it works for him.

In general to workaround profiling rubbish issues you generally `nop` the prefix prior to routing with the code like below. I am not sure whether this code is used anywhere else, but I pretty much faced such issues in the past. Perhaps it is a good idea to put it to Lilu.
```
			auto regWrite = patcher.solveSymbol(index, "__ZN31AppleIntelFramebufferController15WriteRegister32Emj", address, size);
			if (regWrite) {
				auto instCurr = regWrite;
				for (size_t i = 0; i < 5; i++) {
					auto instSize = Disassembler::quickInstructionSize(reinterpret_cast<mach_vm_address_t>(instCurr), 1);
					if (instSize == 0) break; // Unknown instruction
					uint8_t incPrefix[] {0x48, 0xFF, 0x05};
					if (instSize == 7 && !memcmp(reinterpret_cast<void *>(instCurr), incPrefix, sizeof(incPrefix))) {
						auto status = MachInfo::setKernelWriting(true, KernelPatcher::kernelWriteLock);
						if (status == KERN_SUCCESS) {
							for (size_t j = 0; j < 7; j++) reinterpret_cast<uint8_t *>(instCurr)[j] = 0x90; // nop
							MachInfo::setKernelWriting(false, KernelPatcher::kernelWriteLock);
						} else {
							SYSLOG("igfx", "instruction change protection failure %d", status);
						}
					}
					instCurr += instSize;
				}
			}
```

Also, it will be handful if Fraxul moved here.

black-dragon74:
I'll let him know.

Regards

Fraxul:
Thanks for the sample. Don't know why I didn't think of that, but it's a straightforward solution.

With that integrated, I wrote a much simpler version of the patch that only hooks WriteRegister32. I implemented both the CFL-on-CFL driver and CFL-on-KBL driver versions. CFL-on-KBL isn't extensively tested (it's based on code from a config I don't run anymore), but I ran the native CFL version through the wringer with sleep/wake and brightness adjustment -- everything checks out, at least on my test machine. (10.13.6 17G2208 on a 2018 Razer Blade 15" i7-8750H)

Try this on for size:

https://github.com/Fraxul/WhateverGreen/commit/1e92070d4d22a832492f339f05ad814c20efddb1

vit9696:
@Fraxul, well, the changeset does indeed look nice. I would cleanup some parts of it if I were to merge, but generally it is indeed very neat. I wonder why does KBL implementation differ from CFL though. Should not 0xc8254 be restored after sleep in KBL?

As for the rest, I am not sure whether to transition from our current approach to the new one. Ours is no worse in terms of wraps, and I am not positive whether yours is more reliable. Have you tested our current code?

If the new style of patching works more reliably, I guess it makes sense to consider it, but need more data.

Fraxul:
I built master as of this post ( 13a5fbcc8c349f13c7461fa448da1f80b0c3baa4 ) and tested it on my laptop. The brightness control doesn't work properly: it jumps back to low brightness in the middle of the adjustment range, which means that the duty cycle math is overflowing.

I don't remember if I tested sleep/wake for the CFL-on-KBL patch, I'll swap back to my other NVME and do that now.

Fraxul:
CFL-on-KBL sleep/wake display reinit should be fixed in https://github.com/Fraxul/WhateverGreen/commit/b20101e423ed1561d777bb9f9245ddb612a1ea36 .

Fraxul:
@headkaze: yeah, it's weird, I turned on the debug output and verified that the backlight patches are being applied. The overridden `hwSetBacklight` is not being called and I can set a breakpoint in the original `hwSetBacklight` and see this stack trace:

```
AppleIntelCFLGraphicsFramebuffer`AppleIntelFramebufferController::hwSetBacklight(unsigned int)
AppleIntelCFLGraphicsFramebuffer`AppleIntelFramebuffer::setAttributeForConnection(int, unsigned int, unsigned long) + 1602
AppleBacklight`AppleIntelPanelA::doIntegerSet(OSDictionary*, OSSymbol const*, unsigned int) + 840
```

The frequency patch (`0x56ce` -> `0x1d4c0`) was applied, though. Something is clearly not working right on my end, but I haven't figured out what it is yet.


Anyway, my opinion is that patching `WriteRegister32` is the cleanest way to do it: it's unlikely to break with future graphics driver updates and is the minimal surface area (patching a single function instead of all of its callsites). Doing the recomputation of the duty cycle at register write time also lets you grab the PWM frequency from the system, instead of assuming that it's always going to be `0x1d4c0` or requiring the user to provide it with a boot-arg.

headkaze:
You make some good points for the WriteRegister32 patch method. If we can't
get the current method working without patching a bunch more functions then
I agree it does look like a better way.

PR #12, opened 2018-11-19
Add Asus STRIX OC Radeon Vega 64 identifer

Pavo-IM:


vit9696:
Hi Pavo,

We have a policy regarding the names, and vendor-specific ones (like ASUS, Sapphire) are not allowed to be present in the kext for various reasons. See: [the rules](https://github.com/acidanthera/WhateverGreen/blob/master/WhateverGreen/kern_model.cpp#L15-L20).

In case of `Sapphire NITRO+ OC Radeon RX 480` the lower value `Radeon RX 480` will be chosen based on `0xC7` `revision-id`, so no change is necessary. Same with `Asus STRIX OC Radeon RX Vega 64` and `0xC1` Vega `revision-id`.

If one wants a stylised version of the name, he can always provide it via `model` device property.

Cheers,
Vit

PR #15, opened 2019-02-16
Add PowerColor Radeon R9 270 support

nos1609:
Adding preliminary support of [PowerColor R9 270 OC gpu](https://www.techpowerup.com/gpu-specs/powercolor-r9-270-oc.b2629)
screenshot 2019-02-16 at 04 53 53


vit9696:
This ID is not natively supported, and the code males no sense without being referenced.

PR #20, opened 2019-03-31
Set maxLinkRate type to uint32_t so it gets read

bavariancake:
Creating this PR at @0xFireWolf's [suggestion](https://github.com/0xFireWolf/WhateverGreen/pull/1#issuecomment-478317374).

The recent DPCD max link rate patch doesn't read `dpcd-max-link-rate` since `getOSDataValue()` is expecting `uint8_t` instead of `uint32_t` and returns false even when `dpcd-max-link-rate` is present. Fix tested and working.

Let me know if there's a better way to accomplish this. Thanks!


vit9696:
Ah, that was an overlook, sorry. We changed the docs for consistency across all other value, but forgot about to update the variable type. Thanks.

PR #22, opened 2019-05-25
Fix automatic setting of CFG_FB_LIMIT

lwfitzgerald:
Previously the incorrect variable was being used causing `CFG_FB_LIMIT` to
be automatically set to the number of bytes in the `connectors` data
instead of the number of connectors.


vit9696:
Thanks for the patch, should not this be `*sz` instead of `consSize` as `consCount` is not always initialised as mentioned by Clang Analyzer?

lwfitzgerald:
Good point! Fixed!

PR #24, opened 2019-06-22
Added LSPCON driver support and the fix for infinite loop on establishing HDMI connections on IGPU

0xFireWolf:
This pull request contains two fixes.  
  
- Add the driver support for onboard LSPCON chips to enable the DisplayPort to HDMI 2.0 output on some platforms.  
**Brief Description:**  
Recent laptops are equipped with a HDMI 2.0 port now. HDMI 2.0 ports on some laptops are routed to IGPU, but Intel IGPU does not provide native HDMI 2.0 output, so OEMs add a chip named LSPCON on the motherboard to convert DP signals to HDMI 2.0 signals.  
LSPCON chip has two running modes, and this driver automatically configures the adapter to work in the mode that is capable of converting DP to HDMI 2.0 signals, because some adapters are configured in the firmware to work in the HDMI 1.4 mode by default.   
Detailed description could be found [at here in kern_igfx.cpp](https://github.com/0xFireWolf/WhateverGreen/blob/859ee414b86373533c3f702ac45a64c1af654a5a/WhateverGreen/kern_igfx.cpp#L1190)
- Fix the infinite loop when the graphics driver tries to establish a HDMI connection with a higher pixel clock rate, e.g. 533.25 MHz for 4K @ 60Hz.  
**Brief Description:**  
This fix is intended to fix the HDMI 2.0 output issue on my Dell XPS 15 laptop and is now succeeded by the LSPCON driver. However, this fix might still be useful for those who want to have limited 2K/4K experience (i.e. 2K@59Hz or 4K@30Hz) with their HDMI 1.4 port.  
Detailed description could be found [at here in kern_igfx.cpp](https://github.com/0xFireWolf/WhateverGreen/blob/859ee414b86373533c3f702ac45a64c1af654a5a/WhateverGreen/kern_igfx.cpp#L781)  
  
  
Relevant discussion thread: https://www.tonymacx86.com/threads/fix-coffee-lake-intel-uhd-graphics-630-on-macos-mojave-hdmi-output-issue-public-testing-stage.275126/
  
Articles related to this research is still working in progress.  
  
Thanks,  
FireWolf


vit9696:
Hey,

Thanks for this pull-request. Looks quite impressive! It will take me some time to review the code, but all in all it looks reasonably healthy. For the time being

- I noticed that there is some stuff like `OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();`, which is somewhat dangerous to even though in most cases it will likely work just fine.
- `-igfxhdmidivs` is not backed with an IGPU property, perhaps it is worth doing that for parity?

0xFireWolf:
Thanks for your reply. Take your time. :)

> I noticed that there is some stuff like OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();, which is somewhat dangerous to even though in most cases it will likely work just fine.

Yes, in theory it is dangerous, but the property value of `IOFBDependentIndex` always exists and therefore is non-null.

> `-igfxhdmidivs` is not backed with an IGPU property, perhaps it is worth doing that for parity?

Honestly, I was thinking about adding a property for it before.  
Since this fix could be replaced by the LSPCON driver in most cases, I think we don't need to add a property for it. Most people would prefer 4K@60Hz with HDMI 2.0 to 4K@24Hz with HDMI 1.4, so if they really need to apply this fix, they could just add an extra boot-arg.  
For those who use 1080p HDMI display, they don't even need this. That's why Apple's original implementation works with common pixel clock rates under HDMI 1.4 standard.  

However, considering the consistency, we could add a property for it. It doesn't hurt. What's your opinion?

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,584 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: Divider = %d --> P0 = %d; P1 = %d; P2 = %d.\n", context.divider, context.pdiv, context.qdiv, context.kdiv);
+	
+	// Calculate the CRTC parameters
+	uint32_t multiplier = (uint32_t) (context.frequency / 24000000);
+	uint32_t fraction = (uint32_t) (context.frequency - multiplier * 24000000);
+	uint32_t cf15625 = (uint32_t) (context.central / 15625);
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: Multiplier = %d; Fraction = %d; CF15625 = %d.\n", multiplier, fraction, cf15625);
+	auto params = reinterpret_cast<CRTCParams*>(parameters);
+	if (params == nullptr) {
+		DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() Error: Failed to cast the given CRTC parameter pointer.");
```
07151129: This is a bit misleading: reinterpret_cast never fails

WhateverGreen/kern_igfx.hpp
```diff
@@ -457,7 +492,668 @@ class IGFX {
...
+	/**
+	 *	Reflect the `AppleIntelFramebufferController::CRTCParams` struct
+	 *
+	 *	@note Unlike the Intel Linux Graphics Driver,
+	 *	- Apple does not transform the `pdiv`, `qdiv` and `kdiv` fields.
+	 *	- Apple records the final central frequency divided by 15625.
+	 *	@ref static void skl_wrpll_params_populate(params:afe_clock:central_freq:p0:p1:p2:)
+	 *	@seealso https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/gpu/drm/i915/intel_dpll_mgr.c?h=v5.1.13#n1171
+	 */
+	struct CRTCParams
```
07151129: We probably need to pack this struct or use a static_assert to guarantee the correct size when passed to driver (even though it's unlikely to break on X86)

WhateverGreen/kern_igfx.cpp
0xFireWolf: Thanks for your suggestion.

Yes, it is. My intention was to check whether the “parameters” argument is a null pointer or not, but the DBGLOG content is indeed misleading. I will correct it later.

0xFireWolf: Resolved.

WhateverGreen/kern_igfx.cpp
```diff
@@ -141,7 +141,12 @@ void IGFX::init() {
 }
 
 void IGFX::deinit() {
-
+	for (int index = 0; index < arrsize(lspcons); index++) {
+		if (lspcons[index].lspcon != nullptr) {
+			delete lspcons[index].lspcon;
```
vit9696: Please do not use new/delete directly in XNU. Under normal conditions delete must check for nullptr according to C++ standard, but Apple's implementation does not, and thus will crash. See: https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/libkern/c%2B%2B/OSRuntime.cpp#L543

It is very confusing for many people, and thus we invented a concept of `create` and `deleter` methods. Probably it is a good idea to implement them here too, like this:
https://github.com/acidanthera/Lilu/blob/223f20f030d575f9782880e3a7e520fe4dcfc1cd/Lilu/Headers/kern_mach.hpp#L176-L177

WhateverGreen/kern_igfx.cpp
```diff
@@ -160,6 +165,33 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
...
+		// Or if "enable-lspcon-support" is set in IGPU property
+		if (!supportLSPCON)
+			supportLSPCON = info->videoBuiltin->getProperty("enable-lspcon-support") != nullptr;
+		
+		// Read the user-defined IGPU properties to know whether a connector has an onboard LSPCON chip
+		if (supportLSPCON) {
+			char name[48];
+			uint32_t pmode = 0x01; // PCON mode as a fallback value
+			for (size_t index = 0; index < arrsize(lspcons); index++) {
+				bzero(name, 48);
```
vit9696: Use sizeof(name) please.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
 
+void IGFX::populateP0P1P2(struct ProbeContext *context) {
+	uint32_t p = context->divider;
+	uint32_t p0 = 0;
+	uint32_t p1 = 0;
+	uint32_t p2 = 0;
+	
+	// Even divider
+	if (p % 2 == 0) {
+		unsigned int half = p / 2;
```
vit9696: Please use uint32_t for consistency.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	
+	/// All possible central frequency values
+	static constexpr uint64_t centralFrequencies[3] = {8400000000ULL, 9000000000ULL, 9600000000ULL};
+	
+	// Calculate the AFE clock
+	uint64_t afeClock = pixelClock * 5;
+	
+	// Prepare the context for probing P0, P1 and P2
+	struct ProbeContext context;
+	bzero(&context, sizeof(struct ProbeContext));
```
vit9696: Please do not use struct and do not call bzero. Replace it with just ProbeContext context {}; and let the compiler do the job.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	
+	// Apple chooses 400 as the initial minimum deviation
+	// However 400 is too small for a pixel clock like 533.25 MHz (HDMI 2.0 4K @ 60Hz)
+	// Raise the value to UInt64 MAX
+	// It's OK because the deviation is still bound by MAX_POS_DEV and MAX_NEG_DEV.
+	context.minDeviation = UINT64_MAX;
+	
+	for (int dindex = 0; dindex < sizeof(dividers) / sizeof(int); dindex++) {
```
vit9696: Here and in a loop below I am pretty sure it must be not be int (but size_t), and must at least be arrsize(dividers). Most likely, however, you would want to use range-based for once again:

for (auto &divider : dividers)
for (auto &centra :  centralFrequencies)

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: Divider = %d --> P0 = %d; P1 = %d; P2 = %d.\n", context.divider, context.pdiv, context.qdiv, context.kdiv);
+	
+	// Calculate the CRTC parameters
+	uint32_t multiplier = (uint32_t) (context.frequency / 24000000);
+	uint32_t fraction = (uint32_t) (context.frequency - multiplier * 24000000);
+	uint32_t cf15625 = (uint32_t) (context.central / 15625);
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: Multiplier = %d; Fraction = %d; CF15625 = %d.\n", multiplier, fraction, cf15625);
+	// Guard: The given CRTC parameters should never be NULL
+	auto params = reinterpret_cast<CRTCParams*>(parameters);
+	if (params == nullptr) {
```
vit9696: Just as @07151129 mentioned, do not do it like this. Please check parameters first, and then do reinterpret_cast. I.e.

if (parameters == nullptr) {
….
}
auto params = reinterpret_cast<CRTCParams*>(parameters);

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: CTRC parameters have been populated successfully.");
+	return 0;
+}
+
+IGFX::LSPCON::LSPCON(void *controller, IORegistryEntry *framebuffer, void *displayPath) {
+	this->controller = controller;
+	this->framebuffer = framebuffer;
+	this->displayPath = displayPath;
+	this->isActive = false;
+	this->index = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();
```
vit9696: While I am sure it is safe, modern analysers from Xcode 11 will likely error on this. Please check that framebuffer->getProperty("IOFBDependentIndex")) is not nullptr before calling unsigned32BitValue(). E.g.

```C++
auto idxnum  = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"));
if (idxnum != nullptr) {
   this->index = idxnum->unsigned32BitValue();
} else {
  this->index = -1; // or panic or error log, depending on how critical it is
}
```

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+}
+
+IOReturn IGFX::LSPCON::probe() {
+	// Read the adapter info
+	uint8_t buffer[128];
+	bzero(buffer, 128);
```
vit9696:
```
uint8_t buffer[128] {};
```

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	auto info = reinterpret_cast<DisplayPortDualModeAdapterInfo*>(buffer);
+	// Guard: Check whether this is a LSPCON adapter
+	if (!isLSPCONAdapter(info)) {
+		SYSLOG("igfx", "SC: LSPCON::probe() Error: [FB%d] Not a LSPCON DP-HDMI adapter. AdapterID = 0x%02x.", index, info->adapterID);
+		return kIOReturnNotFound;
+	}
+	
+	// Parse the chip vendor
+	char device[8];
+	bzero(device, 8);
```
vit9696: char device[8] {};

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	lilu_os_memcpy(device, info->deviceID, 6);
+	DBGLOG("igfx", "SC: LSPCON::probe() DInfo: [FB%d] Found the LSPCON adapter: %s %s.", index, getVendorString(parseVendor(info)), device);
+	
+	// Parse the current adapter mode
+	Mode mode = parseMode(info->lspconCurrentMode);
+	DBGLOG("igfx", "SC: LSPCON::probe() DInfo: [FB%d] The current adapter mode is %s.", index, getModeString(mode));
+	switch (mode) {
+		case Mode::LevelShifter:
```
vit9696: Does switch really make it more readable? In my opinion, it will be much better with just if…

```C++
if (mode == Mode::ProtocolConverter)
  isActive = true;
else if (mode != Mode::LevelShifter)
 SYSLOG("igfx", "SC: LSPCON::probe() Error: [FB%d] Cannot detect the current adapter mode %d. Assuming Level Shifter mode.", index, mode);
```

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+		default:
+			SYSLOG("igfx", "SC: LSPCON::probe() Error: [FB%d] Cannot detect the current adapter mode. Assuming Level Shifter mode.", index);
+			break;
+	}
+	
+	return kIOReturnSuccess;
+}
+
+IOReturn IGFX::LSPCON::getMode(Mode *mode) {
+	uint8_t byte;
```
vit9696: Maybe move it right above `retVal = callbackIGFX->advReadI2COverAUX`? It is very confusing where it belongs to be honest, especially with the name. I think it may be reasonable to also rename it to hardwareMode or something like that.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+			retVal = callbackIGFX->advReadI2COverAUX(controller, framebuffer, displayPath, DP_DUAL_MODE_ADAPTER_I2C_ADDR, DP_DUAL_MODE_LSPCON_CURRENT_MODE, 1, &byte, 0);
+			
+			// Guard: Can read the current adapter mode successfully
+			if (retVal == kIOReturnSuccess) {
+				DBGLOG("igfx", "SC: LSPCON::getMode() DInfo: [FB%d] The current mode value is 0x%02x.", index, byte);
+				*mode = parseMode(byte);
+				break;
+			}
+			
+			// Sleep 1 ms
```
vit9696: Please explain why you need to sleep for 1 ms. It is relatively clear that IOSleep(1) does that. What is not clear, is why it is needed. Perhaps this comment needs to be removed entirely, and `Try at most 5 times` needs to be more descriptive.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	return retVal;
+}
+
+IOReturn IGFX::LSPCON::setMode(Mode newMode) {
+	// Guard: The given new mode must be valid
+	if (newMode == Mode::Invalid)
+		return kIOReturnAborted;
+	
+	// Guard: Write the new mode
+	uint8_t byte = getModeValue(newMode);
```
vit9696: Again that byte is… confusing. Maybe it should be hardwareMode, or modeRegister, or adapterMode… but not byte.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	if (retVal != kIOReturnSuccess)
+		SYSLOG("igfx", "SC: LSPCON::wakeUpNativeAUX() Error: [FB%d] Failed to wake up the native AUX channel. RV = 0x%llx.", index, retVal);
+	else
+		DBGLOG("igfx", "SC: LSPCON::wakeUpNativeAUX() DInfo: [FB%d] The native AUX channel is up. DPCD Rev = 0x%02x.", index, byte);
+	return retVal;
+}
+
+IOReturn IGFX::wrapReadI2COverAUX(void *that, IORegistryEntry *framebuffer, void *displayPath, uint32_t address, uint16_t length, uint8_t *buffer, bool intermediate, uint8_t flags) {
+	if (callbackIGFX->verboseI2C) {
+		auto index = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();
```
vit9696: Same issue with index as the above.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+		SYSLOG("igfx", "SC:  ReadI2COverAUX() returns 0x%x.", retVal);
+		return retVal;
+	} else {
+		return callbackIGFX->orgReadI2COverAUX(that, framebuffer, displayPath, address, length, buffer, intermediate, flags);
+	}
+}
+
+IOReturn IGFX::wrapWriteI2COverAUX(void *that, IORegistryEntry *framebuffer, void *displayPath, uint32_t address, uint16_t length, uint8_t *buffer, bool intermediate) {
+	if (callbackIGFX->verboseI2C) {
+		auto index = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();
```
vit9696: Same issue with index as the above.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	//    (Laptops/Mobiles start from Skylake platform. e.g. Intel Skull Canyon NUC; Iris Pro 580 and HDMI 2.0)
+	// 2. If your HDMI 2.0 with Intel (U)HD Graphics is working properly, you don't need this fix,
+	//    as the adapter might already be configured in the firmware to run in PCON mode.
+	//
+	// - FireWolf
+	// - 2019.06
+	//
+	
+	// Retrieve the framebuffer index
+	auto index = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"))->unsigned32BitValue();
```
vit9696: Same issue with index as the above.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+		// Already initialized
+		lspcon = framebufferGetLSPCON(index);
+		DBGLOG("igfx", "SC: GetDPCDInfo() DInfo: [FB%d] LSPCON driver (at 0x%llx) has already been initialized for this framebuffer.", index, lspcon);
+		// Confirm that the adapter is running in preferred mode
+		if (lspcon->setModeIfNecessary(pmode) != kIOReturnSuccess) {
+			SYSLOG("igfx", "SC: GetDPCDInfo() Error: [FB%d] The adapter is not running in preferred mode. Failed to update the mode.", index);
+		}
+		// Wake up the native AUX channel if PCON mode is preferred
+		if (pmode == LSPCON::Mode::ProtocolConverter) {
+			if (lspcon->wakeUpNativeAUX() != kIOReturnSuccess)
```
vit9696: Merge that into previous if, currently it indents one level lower for no reason.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+				SYSLOG("igfx", "SC: GetDPCDInfo() Error: [FB%d] Failed to wake up the native AUX channel.", index);
+		}
+		goto org;
+	}
+	
+	// User has specified the existence of onboard LSPCON
+	// Guard: Initialize the driver for this framebuffer
+	lspcon = new LSPCON(that, framebuffer, displayPath);
```
vit9696: Just as mentioned previously, most likely this wants to be LSPCon::create(that, framebuffer, displayPath), same with the delete below.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+		SYSLOG("igfx", "SC: GetDPCDInfo() Error: [FB%d] The adapter is not running in preferred mode. Failed to set the %s mode.", index,  LSPCON::getModeString(pmode));
+	}
+	DBGLOG("igfx", "SC: GetDPCDInfo() DInfo: [FB%d] The adapter is running in preferred mode [%s].", index, LSPCON::getModeString(pmode));
+	
+#ifdef DEBUG
+	framebuffer->setProperty("fw-framebuffer-has-lspcon", framebufferHasLSPCON(index));
+	framebuffer->setProperty("fw-framebuffer-preferred-lspcon-mode", LSPCON::getModeValue(pmode), 8);
+#endif
+	
+org:
```
vit9696: Are you sure this goto makes good sense here? I would say make some inline function in the header if necessary and make the above goto's invoke that function. While goto kind of does make sense for cleanup, here I see no justified reason for its use.

WhateverGreen/kern_igfx.cpp
```diff
@@ -376,6 +408,44 @@ bool IGFX::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t a
 			}
 		}
 		
+		if (hdmiP0P1P2Patch) {
+			auto ppp = patcher.solveSymbol(index, "__ZN31AppleIntelFramebufferController17ComputeHdmiP0P1P2EjP21AppleIntelDisplayPathPNS_10CRTCParamsE", address, size);
+			if (ppp) {
+				patcher.eraseCoverageInstPrefix(ppp);
+				orgComputeHdmiP0P1P2 = reinterpret_cast<decltype(orgComputeHdmiP0P1P2)>(patcher.routeFunction(ppp, reinterpret_cast<mach_vm_address_t>(wrapComputeHdmiP0P1P2), true));
```
vit9696: What is the reason for saving the original function when you do not use it? Most likely you want to just route the function without generating a trampoline. Check routeFunction arguments, it has a bool for this. In this case you will also not need `patcher.eraseCoverageInstPrefix`.

WhateverGreen/kern_igfx.hpp
vit9696: We definitely do not want to pack this, but just as @07151129 static assertions are needed to verify offsets.
Please add code like `static_assert(offsetof(CRTCParams, pdiv) == 0x20), "Invalid pdiv offset, check your compiler");` for critical areas.

WhateverGreen/kern_igfx.hpp
```diff
@@ -457,7 +492,668 @@ class IGFX {
...
+		uint32_t fraction;
+		
+		/// Multiplier of 24 MHz        [`CRTCParams` field offset 0x30]
+		uint32_t multiplier;
+		
+		/// Central Frequency / 15625   [`CRTCParams` field offset 0x34]
+		uint32_t cf15625;
+		
+		/// Other fields that are not of interest
+		uint8_t others[0];
```
vit9696: This is non-standard extension that makes no sense to me. C++ does not support flexible array member, and [0] or [1] are just wrong. In my opinion, remove it or make `[1]`. A comment is most likely more than enough.

WhateverGreen/kern_igfx.hpp
```diff
@@ -457,7 +492,668 @@ class IGFX {
...
+	};
+	
+	/**
+	 *	Represents the current context of probing dividers for HDMI connections
+	 */
+	struct ProbeContext
+	{
```
vit9696: Might you want to put { to `struct ProbeContext`? I.e. `struct ProbeContext {` This is what the rest of the code does. The same issue exists with CRTCParams, DisplayPortDualModeAdapterInfo, and LSPCON.

WhateverGreen/kern_igfx.hpp
```diff
@@ -457,7 +492,668 @@ class IGFX {
...
+		 *
+		 *	@return `kIOReturnSuccess` on success, other errors otherwise.
+		 */
+		IOReturn wakeUpNativeAUX();
+		
+		/**
+		 *	Return `true` if the adapter is active
+		 */
+		inline bool isAdapterActive() {
+			return this->isActive;
```
vit9696: No need for `this` here.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +721,585 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	params->kdiv = context.kdiv;
+	params->multiplier = multiplier;
+	params->fraction = fraction;
+	params->cf15625 = cf15625;
+	DBGLOG("igfx", "SC: ComputeHdmiP0P1P2() DInfo: CTRC parameters have been populated successfully.");
+	return 0;
+}
+
+IGFX::LSPCON::LSPCON(void *controller, IORegistryEntry *framebuffer, void *displayPath) {
+	this->controller = controller;
```
vit9696: Maybe put the constructor to header? It will help to find where members are initialised a lot quicker.

vit9696:
Sorry for the delay, I believe I am done now. I marked the necessary changes inline and mentioned below.

1. For some reason select files got their permissions changed (e.g. 100755 → 100644 for WhateverGreen/kern_igfx.cpp). Was this intentional and what is the right one you think? The change disables writing for anything but the owner, so I get a feeling it may not be desired. If it is a concern, best to address it for all files in a separate patch.

2. I also discovered that Apple-doc comments seem to have tabs after `*`, i.e.

    ```
    /*
     * <tab here> Text
     */
    ```

    The rest of the code uses spaces, so it is probable worth changing it for consistency.

3. `-igfxhdmidivs` I think it probably still makes sense to make it property-controlled. Not much work but better for consistency. Not so long ago I made the NVIDIA property & arg based exactly for that reason.

WhateverGreen/kern_igfx.cpp
0xFireWolf:
```
if (lspcons[index].lspcon != nullptr) {
	delete lspcons[index].lspcon;
}
```
I think I have checked for nullptr before I delete the instance...  

Anyway, I got your idea and have implemented the `creater` and `deleter` methods as suggested, but I have a question related to the implementation of `MachInfo::deleter()`.   Shouldn't we check whether the instance is nullptr or not before deleting it?  Otherwise, what's the point of adding a deleter method that just `delete` the given instance?

```
		static void deleter(LSPCON *instance) {
			if (instance != nullptr)
				delete instance;
		}
```

WhateverGreen/kern_igfx.cpp
0xFireWolf: Ah, sorry about that. I imported the code from my C project, and I forgot to remove it. : )

WhateverGreen/kern_igfx.cpp
0xFireWolf: Thanks for mentioning that. Since those arrays are immutable, I don't need the reference to the element in the array, so I think I could use the value directly.

```
for (auto divider : dividers) { ... }
```

WhateverGreen/kern_igfx.cpp
0xFireWolf: Not at all. Previously, there were just a few lines of code in this function, but later I added functionalities like setting customizable mode, etc. Now I have decided to refactor this method. Please take a look at `framebufferSetupLSPCON()`.

0xFireWolf:
Thanks for reviewing my code. I have pushed a new commit.

>     1. For some reason select files got their permissions changed (e.g. 100755 → 100644 for WhateverGreen/kern_igfx.cpp). Was this intentional and what is the right one you think? The change disables writing for anything but the owner, so I get a feeling it may not be desired. If it is a concern, best to address it for all files in a separate patch.

No, this wasn't intentional. All my private files have permissions 644. I have reverted back to 755 for files in WEG.

> 
>     2. I also discovered that Apple-doc comments seem to have tabs after `*`, i.e.
>        ```
>        /*
>         * <tab here> Text
>         */
>        ```
>        
>        
>        The rest of the code uses spaces, so it is probable worth changing it for consistency.

Done.
 
>     3. `-igfxhdmidivs` I think it probably still makes sense to make it property-controlled. Not much work but better for consistency. Not so long ago I made the NVIDIA property & arg based exactly for that reason.

Done.

0xFireWolf:
By the way, I am considering implementing a helper to retrieve the framebuffer index in a convenient way.  
Something like the following.  

```
struct AppleIntelFramebufferExplorer {
    static bool getIndex(IORegistryEntry *framebuffer, uint32_t &index) {
         auto idxnum = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"));
         if (idxnum != nullptr) {
             index = idxnum->unsigned32bitValue();
             return true;
         } else {
             return false;
         }
    }
}
```

What's your suggestion?

WhateverGreen/kern_igfx.cpp
vit9696: No, it should definitely be caller-dependent. The original logic was obscure due to no non-null attribute and sometimes inconsistent code. I fixed this in https://github.com/acidanthera/Lilu/commit/f45255eb785a2a2728bb07815e50cb8219bfeb31.

vit9696:
> By the way, I am considering implementing a helper to retrieve the framebuffer index in a convenient way.

Something like that could do, as you see fit I guess.

vit9696:
All in all this looks quite good to me aside the small nuances I added. Please fix them up and I will merge as soon as I can. Thank you so much for this awesome stuff!

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +720,579 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	// HDMI 1.4 port, you might find this fix helpful.
+	//
+	// For those who have a laptop or PC with HDMI 2.0 routed to IGPU and have HDMI output issues,
+	// it is still recommended to enable the LSPCON driver support to have full HDMI 2.0 experience.
+	//
+	// - FireWolf
+	// - 2019.06
+	//
+	
+	static_assert(__offsetof(CRTCParams, pdiv) == 0x20, "Invalid pdiv offset, please check your compiler.");
```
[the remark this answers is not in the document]
0xFireWolf: Ah, my bad. It's auto-correction done by Xcode. I should have noticed that.

sc

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +720,579 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	else
+		DBGLOG("igfx", "SC: LSPCON::wakeUpNativeAUX() DInfo: [FB%d] The native AUX channel is up. DPCD Rev = 0x%02x.", index, byte);
+	return retVal;
+}
+
+IOReturn IGFX::wrapReadI2COverAUX(void *that, IORegistryEntry *framebuffer, void *displayPath, uint32_t address, uint16_t length, uint8_t *buffer, bool intermediate, uint8_t flags) {
+	if (callbackIGFX->verboseI2C) {
+		auto idxnum = OSDynamicCast(OSNumber, framebuffer->getProperty("IOFBDependentIndex"));
+		auto index = idxnum != nullptr ? idxnum->unsigned32BitValue() : -1;
+		//->unsigned32BitValue();
```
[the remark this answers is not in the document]
0xFireWolf: Did not generate a warning, but I have replaced it with the `getIndex()` helper.

WhateverGreen/kern_igfx.cpp
0xFireWolf: Got it. :D

0xFireWolf:
> All in all this looks quite good to me aside the small nuances I added. Please fix them up and I will merge as soon as I can. Thank you so much for this awesome stuff!

I have pushed new commits to the repo. Thanks again for reviewing my code.

WhateverGreen/kern_igfx.cpp
```diff
@@ -651,6 +719,574 @@ int IGFX::wrapReadAUX(void *that, IORegistryEntry *framebuffer, uint32_t address
...
+	else
+		DBGLOG("igfx", "SC: LSPCON::wakeUpNativeAUX() DInfo: [FB%d] The native AUX channel is up. DPCD Rev = 0x%02x.", index, byte);
+	return retVal;
+}
+
+IOReturn IGFX::wrapReadI2COverAUX(void *that, IORegistryEntry *framebuffer, void *displayPath, uint32_t address, uint16_t length, uint8_t *buffer, bool intermediate, uint8_t flags) {
+	if (callbackIGFX->verboseI2C) {
+		uint32_t index = 0xFF;
+		AppleIntelFramebufferExplorer::getIndex(framebuffer, index);
+		SYSLOG("igfx", "SC:  ReadI2COverAUX() called. FB%d: Addr = 0x%02x; Len = %02d; MOT = %d; Flags = %d.",
```
vit9696: Ew, sorry, did not see that previously. I am pretty sure it must be DBGLOG. Here and below, also in wrapWriteI2COverAUX

0xFireWolf: Hmm, `verboseI2C` is enabled by `-igfxi2cdbg`, so if an user has added this boot argument, I think user should always see this line even though he is using the RELEASE version. That's why I choose SYSLOG.

0xFireWolf: Wait, it makes more sense to use the DEBUG version with this boot argument, I will fix it now.

0xFireWolf: Done.

vit9696:
Thank you! Merged!

PR #27, opened 2019-07-06
LSPCON: Fix an issue that the driver fails to set the mode after the adapter power is off

0xFireWolf:
LSPCON adapters on some laptops are automatically switched back to LS mode on power off.  
  
This fix removes the property `LSPCON::isActive` so that `LSPCON::isRunningInMode()` method always reads the latest mode from the adapter directly instead of the "cached" current mode via `isActive`.  
  
This solves the black screen issue of HDMI 2.0 output on laptops such as Dell XPS 15 9570 after the system wakes up.


PMheart:
Thanks! Merged.

PR #29, opened 2019-07-19
igfx: Implement manual override and auto built-in disable

07151129:


bgman111111:
make intel first generation for lenovo t410. If you need something comment and i will extract it. you have to comment how to be done.

vit9696:
Thanks =)

PR #31, opened 2019-08-02
AMD GPU: show real model string when spoofing

osy86:
It is common to spoof a card's device-id. In that case, the model string displayed should reflect the true card's string instead. We look at the PCI config registers for the real device id instead of the IORegistry.

### DIFF
--- a/WhateverGreen/kern_model.cpp
+++ b/WhateverGreen/kern_model.cpp
@@ -741,6 +741,18 @@ static constexpr Model dev6939[] {
 	{Model::DetectDef, 0x0000, 0x0000, 0x0000, "AMD Radeon R9 285"}
 };
 
+static constexpr Model dev694c[] {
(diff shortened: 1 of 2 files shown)


vandroiy2013:
Hi
kern_model.cpp is only for native cards. 
Does Vega M ID's present in Apple AMD kexts&

osy86:
Oh, I see. No, Vega M is not in Apple kext, we have to spoof a different ID for the drivers to load. The problem is that once spoofed, OSX shows the wrong model string if WG is also loaded.

vandroiy2013:
You have to insert model with device properties. Read the manual.

osy86:
Sure, but isn't it better to detect it automatically?

vandroiy2013:
No. Because "automatic detection" used only for native cards. If you spoof device-id then it is not hard for you to write correct name in device properties.

osy86:
Okay sounds good

PR #33, opened 2019-08-23
Changed identifier for 67DF / 00CF

BenTalagan:
Hi, I have an RX 570 armor 8GO that identifies with 0x67DF, 0x00CF. It's currently marked as Radeon RX 470 in MacOS.

Seems like it's a common thing to have identifiers shared between Radeon 4XX and 5XX. So I have copied your way of handling this from latest commit 6d4920148feb51ef2e92e36aa42e9a77d68b4d73 to fix the identification.

Hope this helps!

### DIFF
--- a/WhateverGreen/kern_model.cpp
+++ b/WhateverGreen/kern_model.cpp
@@ -407,7 +407,7 @@ static constexpr Model dev67df[] {
 	{Model::DetectRev, 0x0000, 0x0000, 0x00c5, "Radeon RX 470"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00c6, "Radeon RX 570"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00c7, "Radeon RX 480"},
-	{Model::DetectRev, 0x0000, 0x0000, 0x00cf, "Radeon RX 470"},
+	{Model::DetectRev, 0x0000, 0x0000, 0x00cf, "Radeon RX 470/570"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00d7, "Radeon RX 470"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00e0, "Radeon RX 470"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00e7, "Radeon RX 580"},
--- a/WhateverGreen/kern_model.cpp
+++ b/WhateverGreen/kern_model.cpp
@@ -399,6 +399,7 @@ static constexpr Model dev67c7[] {
 };
 
 static constexpr Model dev67df[] {
+  {Model::DetectSub, 0x1462, 0x341E, 0x0000, "Radeon RX 570 Armor"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00e1, "Radeon RX 590"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00c1, "Radeon RX 580"},
 	{Model::DetectRev, 0x0000, 0x0000, 0x00c2, "Radeon RX 570"},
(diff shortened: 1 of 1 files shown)


vit9696:
Thanks for the patch. In my opinion, it is best to not only change 470 to 470/570, but also add an extra entry with your `subsystem-vendor-id` and `subsystem-device-id` (see your IOReg for the values), so that your card displays as 570 and others show as 470/570.

BenTalagan:
Ok, `ioreg -lbs` gave me the following info : 

```
- "compatible" = <"pci1462,341e","pci1002,67df","pciclass,030000","PEGP","GFX0">
- "vendor-id" = <02100000>
- "device-id" = <df670000>
- "subsystem-id" = <1e340000>
- "subsystem-vendor-id" = <62140000>
```
Thus : 
				  
- Vendor ID : 		0x1002
- Device ID : 		0x67DF 		 
- Sub Vendor ID : 0x1462
- Sub Device ID : 0x341E

So I have added the card accordingly, though I was not sure if Model::DetectSub, Model::DetectRev, Model::DetectDef could be mixed all 3 in the same struct since there are no other examples in the file, but the code seems to give priority to Sub then Rev so I thought it was ok to do it like this. I have compiled WG to test it and the identification works. Just tell me if it's ok.

Note : Since [https://pci-ids.ucw.cz/read/PC/1002/67df/1462341e](https://pci-ids.ucw.cz/read/PC/1002/67df/1462341e) gives the complete name with "570 Armor" I used that categorisation too (though it can either be 4 or 8 GB).

vandroiy2013:
No need to write Armor.

vandroiy2013:
Also you need to specify revision

vandroiy2013:
This new line must be one line up the line you fixed firs time!

vandroiy2013:
Added manually https://github.com/acidanthera/WhateverGreen/commit/6ba30e9d20a0d96c4bad7f4a0379e76275afa7e9

BenTalagan:
Hi, are you really sure your 6ba30e9d20a0d96c4bad7f4a0379e76275afa7e9 commit is ok? The line I had added on the second time was for a sub `Model::DetectSub` not for a revision `Model::DetectRev`. Seems to me that your correction will have no effect in the `getRadeonModel` function.

BenTalagan:
Also, this sub product is **really** a 570 **Armor** (so it seems to me that, in a context of a sub it's not shocking to have the precision, on the contrary).

BenTalagan:
It seems to me that your last week commit 0615d796346f5512596ec9ded385862b89af234e suffers also the same problem.

BenTalagan:
@vandroiy2013 : after reading again and again the `getRadeonModel` function, it seems to me that what you want for 6ba30e9d20a0d96c4bad7f4a0379e76275afa7e9 and 0615d796346f5512596ec9ded385862b89af234e is `Model::DetectAll`, not `Model::DetectRev`, right (or it just does not make sense to pass sub and subven parameters ?

vit9696:
Actually it looks like we messed up badly with these. Looks like I myself forgot how it works and borked it even further in https://github.com/acidanthera/WhateverGreen/commit/6d4920148feb51ef2e92e36aa42e9a77d68b4d73. Fixed both commits in https://github.com/acidanthera/WhateverGreen/commit/81ef11649983b859460e61c1c78353581b5cc214. Thanks!

As for `Armor`, we do not use vendor suffixes in any models, maybe it is not nice, but that's the design decision we follow.

BenTalagan:
> As for Armor, we do not use vendor suffixes in any models, maybe it is not nice, but that's the design decision we follow.

Perfectly clear and understandable. Thanks!

> Actually it looks like we messed up badly with these. Looks like I myself forgot how it works and borked it even further in 6d49201. Fixed both commits in 81ef116. Thanks!

Awesome. Thanks for the fix, I'm glad it helped!

PR #38, opened 2020-01-01
Adds Comet Lake laptop device ID.

fewtarius:
This works perfectly on my Comet Lake laptop.

### DIFF
--- a/Manual/SSDT-PNLF.dsl
+++ b/Manual/SSDT-PNLF.dsl
@@ -177,6 +177,8 @@ DefinitionBlock("", "SSDT", 2, "ACDT", "PNLF", 0)
             {
                 // CoffeeLake identifiers from AppleIntelCFLGraphicsFramebuffer.kext
                 0x3e9b, 0x3ea5, 0x3e92, 0x3e91,
+                // Canon Lake
+                0x9B41,
             }, MEQ, Local0, MTR, 0, 0))
             {
                 if (Ones == Local2) { Local2 = COFFEELAKE_PWMMAX }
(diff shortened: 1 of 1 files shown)


Andrey1970AppleLife:
It doesn't make sense.
AppleIntelCFLGraphicsFramebuffer.kext doesn't contain device-id 0x9B41
You shall use a fake device-id as CFL.

fewtarius:
Sorry to bring this one back out of the blue but I'm just finishing up this build.  I'm already using a fake device ID of 9B3E0000 which does not work with the unpatched PNLF, nor does it work with that plus the cfl brightness WEG properties patch.  The only way brightness works using the non-modified pnlf ssdt is to use device id C0870000 (UHD 617) but using that device id along with igfxfw=2 causes a sleep panic.  Using 9B3E0000 + igfxfw=2 and this pnlf adjustment, everything works perfectly.  Should I file a bug report instead?

Andrey1970AppleLife:
Use SSDT-PNLFCFL.aml
https://github.com/acidanthera/WhateverGreen/blob/master/Manual/FAQ.IntelHD.en.md

andrewwyatt:
> Use SSDT-PNLFCFL.aml
> https://github.com/acidanthera/WhateverGreen/blob/master/Manual/FAQ.IntelHD.en.md

Thanks, this fixed the problem for me.  I thought the device property did the same thing.

fewtarius:
Thanks, this fixed it.  I mistakenly thought that the device property patch did the same thing.

PR #39, opened 2020-01-10
Add codec hardware info spoofing for AMD VCE

osy86:
When an AMD card is PID spoofed, video encoding does not work because in
AMDRadeonVADriver, `VAAcceleratorInfo::identify` gets the PID by calling
up to AMDRadeonX4000.kext (in a method
`AMDRadeonX4000_AMDAccelVideoContext::getHWInfo` which gets the PID from
`AMDRadeonX4000_AMDHardware::getHWInfo` which is set up by
`AMDRadeonX4000_AMDHardware::initHWInfo` which copies the value from
`AMDRadeonX4000_AMDHardware::init` which reads the PCI registers).

We patch `getHWInfo` to return the spoofed PID (from `device-id`) and this
way, when AppleGVA queries AMDRadeonVADriver, it will no longer return an
error.

Additionally, Shiki patches can be applied to get DRM to work.

### DIFF
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -107,6 +107,9 @@ void RAD::init() {
 
 	// Broken drivers can still let us boot in vesa mode
 	forceVesaMode = checkKernelArgument("-radvesa");
+	
+	// Fix codec PID to be spoofed PID if requested
+	forceCodecInfo = checkKernelArgument("-radcodec");
 
 	// To support overriding connectors and -radvesa mode we need to patch AMDSupport.
 	lilu.onKextLoadForce(&kextRadeonSupport);
@@ -408,6 +411,12 @@ void RAD::processHardwareKext(KernelPatcher &patcher, size_t hwIndex, mach_vm_ad
 			patcher.applyLookupPatch(&p);
 			patcher.clearError();
 		}
+	}
+
+	// Patch AppleGVA support for non-supported models
+	if (forceCodecInfo) {
+		KernelPatcher::RouteRequest request(getHWInfoProcNames[hwIndex], wrapGetHWInfo[hwIndex], orgGetHWInfo[hwIndex]);
+		patcher.routeMultiple(hardware.loadIndex, &request, 1, address, size);
 	}
 }
 
@@ -1080,4 +1089,23 @@ bool RAD::wrapNotifyLinkChange(void *atiDeviceControl, kAGDCRegisterLinkControlE
 	}
 
 	return ret;
+}
+
+void RAD::updateGetHWInfo(IOService *accelVideoCtx, void *hwInfo) {
+	IOService *accel, *pciDev;
+	accel = OSDynamicCast(IOService, accelVideoCtx->getParentEntry(gIOServicePlane));
+	if (accel == NULL) {
+		SYSLOG("rad", "getHWInfo: no parent found for accelVideoCtx!");
+		return;
+	}
+	pciDev = OSDynamicCast(IOService, accel->getParentEntry(gIOServicePlane));
+	if (pciDev == NULL) {
+		SYSLOG("rad", "getHWInfo: no parent found for accel!");
+		return;
+	}
+	uint16_t org = *(uint16_t *)((char *)hwInfo + 0x4);
+	uint32_t dev;
+	WIOKit::getOSDataValue(pciDev, "device-id", dev);
+	DBGLOG("rad", "getHWInfo: original PID: 0x%04X, replaced PID: 0x%04X", org, dev);
+	*(uint16_t *)((char *)hwInfo + 0x4) = dev;
 }
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -1110,5 +1110,5 @@ void RAD::updateGetHWInfo(IOService *accelVideoCtx, void *hwInfo) {
 		WIOKit::getOSDataValue(pciDev, "device-id", dev);
 	}
 	DBGLOG("rad", "getHWInfo: original PID: 0x%04X, replaced PID: 0x%04X", org, dev);
-	org = (uint16_t)dev;
+	org = static_cast<uint16_t>(dev);
 }
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -240,7 +240,7 @@ void RAD::initHardwareKextMods() {
 			getFrameBufferProcNames[IndexRadeonHardwareX4000][i] = nullptr;
 
 		// We have nothing to do for these kexts on recent systems
-		if (!fixConfigName && !forceOpenGL) {
+		if (!fixConfigName && !forceOpenGL && !forceCodecInfo) {
 			// X4000 kext is not included in this list as we need to fix GVA properties for most of its GPUs
 			kextRadeonHardware[IndexRadeonHardwareX5000].switchOff();
 			kextRadeonHardware[IndexRadeonHardwareX6000].switchOff();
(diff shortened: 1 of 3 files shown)


WhateverGreen/kern_rad.cpp
```diff
@@ -107,6 +107,9 @@ void RAD::init() {
 
 	// Broken drivers can still let us boot in vesa mode
 	forceVesaMode = checkKernelArgument("-radvesa");
+	
+	// Fix codec PID to be spoofed PID if requested
+	forceCodecInfo = checkKernelArgument("-radcodec");
```
vit9696: If we decide that this needs a boot argument, I request us also add a DeviceProperty with the same purpose.

WhateverGreen/kern_rad.cpp
```diff
@@ -107,6 +107,9 @@ void RAD::init() {
...
```
vit9696: You also need to update initHardwareKextMods. Find `X4000 kext is not included in this list as we need to fix GVA properties for most of its GPUs`. At first we could change the condition to include `&& !forceCodecInfo`, but that will not work if we add a DeviceProperty. Most likely the decision needs to be split:

- Disable incompatible kexts (missing in the target system) in initHardwareKextMods.
- Disable unused kexts (not required by the configuration) in some other method called from processKernel.

WhateverGreen/kern_rad.cpp
```diff
@@ -409,6 +412,12 @@ void RAD::processHardwareKext(KernelPatcher &patcher, size_t hwIndex, mach_vm_ad
 			patcher.clearError();
 		}
 	}
+
+	// Patch AppleGVA support for non-supported models
+	if (forceCodecInfo) {
+		KernelPatcher::RouteRequest request(getHWInfoProcNames[hwIndex], wrapGetHWInfo[hwIndex], orgGetHWInfo[hwIndex]);
```
vit9696: I believe this one is missing a `&& getHWInfoProcNames[hwIndex] != nullptr` check.

WhateverGreen/kern_rad.cpp
```diff
@@ -1081,3 +1090,22 @@ bool RAD::wrapNotifyLinkChange(void *atiDeviceControl, kAGDCRegisterLinkControlE
...
+	if (accel == NULL) {
+		SYSLOG("rad", "getHWInfo: no parent found for accelVideoCtx!");
+		return;
+	}
+	pciDev = OSDynamicCast(IOService, accel->getParentEntry(gIOServicePlane));
+	if (pciDev == NULL) {
+		SYSLOG("rad", "getHWInfo: no parent found for accel!");
+		return;
+	}
+	uint16_t org = *(uint16_t *)((char *)hwInfo + 0x4);
```
vit9696: Please use `getMember` from `kern_util.hpp` here and below. `getMember<uint16_t>(hwInfo, 0x4)` will be a lot more readable.

WhateverGreen/kern_rad.hpp
```diff
@@ -269,6 +274,43 @@ class RAD {
...
+	/**
+	 *  getHWInfo function type
+	 */
+	using t_getHWInfo = int (*)(IOService *accelVideoCtx, void *hwInfo);
+
+	/**
+	 *  getHWInfo wrapping functions used for AppleGVA enable patch
+	 */
+	template <size_t Index>
+	static int populateGetHWInfo(IOService *accelVideoCtx, void *hwInfo) {
```
vit9696: I believe it is not int but rather IOReturn, given the scope.

vit9696:
In fact, I think we want two more changes to this patch:

- Make device-id hooking firstly try using codec-device-id, and then fallback to device-id, as there may be ids without VCE support but with acceleration support.
- Include X6000 hardware as well, as the change is likely required there too.

WhateverGreen/kern_rad.hpp
07151129: It is IOReturn:

```
__int64 __fastcall AMDRadeonX4000_AMDAccelVideoContext::getHWInfo(__int64 a1, sHardwareInfo *a2)
{
...
if ( !a2 )
    return 0xE00002C2LL;
```

WhateverGreen/kern_rad.cpp
07151129: You should also try and compute struct member offsets in runtime rather than hardcode them for better compatibility (see https://github.com/acidanthera/NVMeFix/blob/master/NVMeFix/NVMeFix.cpp#L76, https://github.com/acidanthera/NVMeFix/blob/master/NVMeFix/NVMeFixPlugin.hpp#L159 for example.) Of course, this only makes sense to do when you believe the code accessing the relevant struct member is less volatile than the struct definition itself, and the mov is not that hard to find, etc.

WhateverGreen/kern_rad.cpp
osy86: Any reason why `-radvesa` `-rad24` and `-raddvi` do not have DeviceProperty? I used those as a template.

WhateverGreen/kern_rad.cpp
osy86: The struct should be stable as it's the second member of a pretty large struct. The contents themselves defines all sorts of different values for various GPUs so I think it's harder.

osy86:
I can add X6000 but note that I do not have X5000 or X6000 to test with.

If we were to add `force-codec` as a property, then we would have to make it per-card (in case of multiple AMD cards). Whereas right now using `-radcodec` enables it globally.

We can either make it global if ANY card enables it (unsure if this would cause issues if say, a system has 1 compatible and 1 incompatible card) or we keep track of `force-codec` per card and flag it that way. This in addition to the above mentioned issue with `initHardwareKextMods` makes me think it may not be worth doing.

WhateverGreen/kern_rad.cpp
vit9696: -radvesa is a way to boot a failing system. But the others are bugs or rather they were added before we started to implement property mirroring. We should fix them sooner or later. I guess I could handle that myself and also fix your -radcodec.

WhateverGreen/kern_rad.hpp
```diff
@@ -269,6 +274,44 @@ class RAD {
...
+		}
+		return 1;
+	}
+
+	/**
+	 *  Wrapped getHWInfo functions
+	 */
+	t_getHWInfo wrapGetHWInfo[MaxRadeonHardware] {
+		[RAD::IndexRadeonHardwareX4000] = populateGetHWInfo<RAD::IndexRadeonHardwareX4000>,
+		[RAD::IndexRadeonHardwareX5000] = populateGetHWInfo<RAD::IndexRadeonHardwareX5000>
```
vit9696: An entry is missing here for X6000.

WhateverGreen/kern_rad.cpp
```diff
@@ -1081,3 +1090,25 @@ bool RAD::wrapNotifyLinkChange(void *atiDeviceControl, kAGDCRegisterLinkControlE
...
+		return;
+	}
+	uint16_t &org = getMember<uint16_t>(hwInfo, 0x4);
+	uint32_t dev = org;
+	if (!WIOKit::getOSDataValue(pciDev, "codec-device-id", dev)) {
+		// fallback to device-id only if we do not have codec-device-id
+		WIOKit::getOSDataValue(pciDev, "device-id", dev);
+	}
+	DBGLOG("rad", "getHWInfo: original PID: 0x%04X, replaced PID: 0x%04X", org, dev);
+	org = (uint16_t)dev;
```
vit9696: Maybe make this static_cast for consistency with the other code?

WhateverGreen/kern_rad.hpp
```diff
@@ -269,6 +274,44 @@ class RAD {
...
+	template <size_t Index>
+	static IOReturn populateGetHWInfo(IOService *accelVideoCtx, void *hwInfo) {
+		if (callbackRAD->orgGetHWInfo[Index]) {
+			int ret = FunctionCast(populateGetHWInfo<Index>, callbackRAD->orgGetHWInfo[Index])(accelVideoCtx, hwInfo);
+			callbackRAD->updateGetHWInfo(accelVideoCtx, hwInfo);
+			return ret;
+		} else {
+			SYSLOG("rad", "populateGetHWInfo invalid use for %lu", Index);
+		}
+		return 1;
```
vit9696: Maybe change this to kIOReturnInvalid or something alike?

vit9696:
Thanks, merged

PR #40, opened 2020-01-18
Update kern_model.cpp

wiblade:
Complete the names of some AMD Graphics model(Radeon-AMD Radeon)


vit9696:
The names you changed follow the convention used in real Macs. Did that change? If not, I am not sure we should follow.

PR #41, opened 2020-02-10
Extended maximum link rate fix for new laptops

0xFireWolf:
#### Abstract:  
Previously, the maximum link rate fix has addressed the issue that an invalid link rate value is returned from the DPCD receiver capabilities buffer of the built-in display. However, some displays on new laptops, such as Dell Inspiron 7590 with Sharp display, have defined extended DPCD receiver capabilities and these fields reside at DPCD address space `0x2200`. As a result, the graphics driver reads the invalid maximum link rate value from 0x2200 and overrides the one fixed by WhateverGreen. This improved fix now also corrects the value in the buffer read from extended DPCD.

#### Relevant Discussion:
https://github.com/tctien342/Dell-Inspiron-7591-Hackintosh/issues/6

Thank you.

FireWolf
2020.02.10


vit9696:
Thanks!

PR #43, opened 2020-04-24
Add wegtree=0/1 boot argument and rebuild-device-tree property

cdf:
For property mirroring. Note that -wegtree remains as an alias.


vit9696:
Reworked a bit, but otherwise good. Will merge once CI finishes. Thank you!

PR #45, opened 2020-06-16
Add option to force enable Metal for offline iGPU

osy86:
By default, `IntelAccelerator::start` will remove the property `MetalPluginName` if the number of
framebuffers is 0. On Macs and Hackintoshes alike, this means that an iGPU with no connectors cannot
be used for Metal's compute APIs.

This new option will re-add `MetalPluginName` in order to allow Metal to be used.

### DIFF
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -310,6 +310,10 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
 		int gl = info->videoBuiltin->getProperty("disable-metal") != nullptr;
 		PE_parse_boot_argn("igfxgl", &gl, sizeof(gl));
 		forceOpenGL = gl == 1;
+		
+		int metal = info->videoBuiltin->getProperty("enable-metal") != nullptr;
+		PE_parse_boot_argn("igfxmetal", &metal, sizeof(metal));
+		forceMetal = metal == 1;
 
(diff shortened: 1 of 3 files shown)


osy86:
Reference: https://github.com/osy86/HaC-Mini/issues/213

WhateverGreen/kern_igfx.cpp
```diff
@@ -761,6 +765,9 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
 
 		}
 	}
+	
+	OSObject *metalPluginName = that->getProperty("MetalPluginName");
```
vit9696, line 769: I guess you should dynamic-cast here, as doing it later may result in null pointer dereference. Also, please check the CI failure.

osy86: `getProperty` returns an `OSObject *` so I think it's fine. But I should add `retain` and `release` because it's possible that `start` will dealloc the property.

vit9696: The issue was that you dunamic casted a pointer and immediately dereferenced it by getting C string without checking for NULL. The latest change resolved the problem, but made clang static analyzer unhappy: https://travis-ci.com/github/acidanthera/WhateverGreen/jobs/350020803

Could you please fix it up?

osy86: I think the static analyzer is too stupid to recognize that callbackIGFX->forceMetal cannot change between calls. I can try to always retain and release it but it might look ugly.

PR #49, opened 2020-07-03
Port i915 force wake workaround

07151129:
KBL/CFL only for now. The code almost exactly copies `IntelAccelerator::SafeForceWakeMultithreaded` logic, but does the fallback trick at first timeout.

Most users don't experience ForceWake timeouts, but this should fix issues like https://github.com/acidanthera/bugtracker/issues/1004.


zhen-zen:
Maybe this PR also fixes https://github.com/acidanthera/bugtracker/issues/967 ?

PR #62, opened 2020-08-23
Support all valid Core Display Clock frequencies to fix the kernel panic on Intel ICL platforms

0xFireWolf:
#### Abstract

Core Display Clock (CDCLK) is one of the primary clocks used by the display engine to do its work. Apple's graphics driver expects that the firmware has already set the clock frequency to either 652.8 MHz or 648 MHz, but quite a few laptops set it to a much lower value, such as 172.8 MHz, and hence a kernel panic of "Unsupported CD clock decimal frequency" occurs to indicate the precondition failure.  

Since the graphics driver relies on the above assumption, this patch reprograms the clock and sets its frequency to a supported value based on the hardware reference clock, so that the kernel panic can be avoided and the built-in display can be lit up with acceleration successfully. 

**Original Research:**
https://www.firewolf.science/2020/08/ice-lake-intel-iris-plus-graphics-on-macos-catalina-a-solution-to-the-kernel-panic-due-to-unsupported-core-display-clock-frequencies-in-the-framebuffer-driver/

Thank you.

-- FireWolf

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ WhateverGreen
 - Fixes choppy video playback on Intel Kaby Lake and newer.
 - Fixes black screen on Intel HD since 10.15.5.
 - Adds workaround for rare force wake timeout panics on Intel KBL and CFL.
-- Support all valid Core Display Clock (CDCLK) freqencies on Intel ICL platforms.
+- Supports all valid Core Display Clock (CDCLK) freqencies on Intel ICL platforms.
 
 #### Documentation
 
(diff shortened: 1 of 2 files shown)


andreszerocross:
image

I am still get Kernel Panic but not about CDC again,, I have built Latest Commit Lilu.kext and WhateverGreen kext and add "enable-dpcd-max-link-rate-fix" property in DSDT.

Any clue???

PR #64, opened 2020-08-31
Fix the kernel panic caused by an incorrectly calculated amount of DVMT pre-allocated memory on Intel ICL platforms

0xFireWolf:
**Abstract:**

Apple’s graphics driver reads the DVMT value set in the BIOS or the firmware and uses a “magic” formula to calculate the amount of memory in bytes. Unfortunately, the formula only works for a pre-allocated memory size that is a multiple of 32MB. Non-Apple laptops normally set DVMT to 32MB on CFL- platforms, and related kernel panics can be solved by patching the framebuffer properly.  

However, problem arises as laptops now have DVMT set to 60MB on ICL+ platforms by default, and the framebuffer controller ends up with initializing the stolen memory manager with an incorrect amount of pre-allocated memory. Later, the accelerator driver fails to read the SKU number and reports a kernel panic of an unsupported ICL SKU. We could manually fix this issue, but then the accelerator cannot allocate objects on the stolen memory area, and hence another kernel panic is triggered inside the `IGAccelTask::withOptions()` function.

Even though one might be able to modify DVMT settings via `EFI shell` or `RU.EFI`, these methods are not applicable to some laptops, such as Surface Pro 7, that use custom firmware or have a strict BIOS protection. As such, this patch locates instructions related to calculating the amount of memory available and identifies which register stores the final result. It calculates the correct number of bytes beforehand and finds the proper instruction to copy the value to the result register. Consequently, the framebuffer controller will initialize the memory manager with proper values and aforementioned kernel panics can be avoided.

**Future Maintainability:**

As mentioned in the last pull request, `IGFX` contains massive code and is getting hard to work with. In order to ease the development of new patches, I would like to propose the following changes.

My intention is to separate the giant IGFX module into many tiny submodules. I extract an interface `PatchSubmodule` to define a submodule for a patch, and it would be easier for IGFX to interact with each submodule. Ideally, IGFX maintains a collection of submodules, and when it processes the kernel, the request will be redirected to the `processKernel()` function implemented by each submodule. (same for `processKext()`)

As such, each submodule checks the boot argument to determine whether it should be enabled and routes functions if necessary, and we no longer need to keep adding huge blocks of code to IGFX’s `process*()` functions. Additionally, IGFX uses the public properties, `enabled`, `requiresPatchingFramebuffer` and `requiresPatchingGraphics` to determine whether it should enable the submodule and disable kext patching or not.

DVMT calculation fix is implemented as a separate submodule, and you could see how it is used by IGFX in my comments. Let me know if this is acceptable.

Thank you.

-- FireWolf


vit9696:
Hi, thx for the changes. Could you explain why simply patching the framebuffer to require less DVMT memory fails? Is it the calculation itself in the driver that is wrong? If so, could you please provide more details on it?

0xFireWolf:
> Hi, thx for the changes. Could you explain why simply patching the framebuffer to require less DVMT memory fails? Is it the calculation itself in the driver that is wrong? If so, could you please provide more details on it?

The graphics memory size (GMS) can be found in PCI0,2,0 at offset 0x50. The high 8 bits store the DVMT pre-allocated memory value set in BIOS. For example, 0x01, 0x02, ..., 0x10 represents 32MB, 64MB, ..., 512MB, while 0xF0, 0xF1, ... 0xFE represents 4MB, 8MB, ..., 60MB.

You could find all possible DVMT values from Section 4.1.28 in Intel's manual at [here](https://www.intel.com/content/www/us/en/products/docs/processors/core/10th-gen-core-families-datasheet-vol-2.html).

Apple uses a "magic" formula `(GGCRegisterValue << 0x11) & 0xFE000000` to calculate the size of pre-allocated memory in bytes, where `GGCRegisterValue` is the 16-bit integer read from 0x50. For example, if the value is `0x01FF` (i.e DVMT = 32MB and we don't care about low 8 bits), the output of the above formula is `(0x01FF << 0x11) & 0xFE000000 = 0x2000000 = 33554432 Bytes = 32 MB`. However, if the value is `0xFEFF` (i.e. DVMT = 60MB), the output becomes 0. 

Previously, laptops have DVMT set to 32M on CFL- platforms, so this calculation is not an issue, and we only need to patch the framebuffer. Nowadays, Ice Lake-based laptops have DVMT set to 60M instead, so the calculation in the driver becomes problematic.

Hope this could clarify the problem. 

-- FireWolf

WhateverGreen/kern_igfx.cpp
```diff
@@ -593,6 +605,10 @@ bool IGFX::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t a
 			}
 		}
 
+		// We could iterate through each submodule and redirect the request if and only if the submodule is enabled
+		if (modDVMTCalcFix.enabled)
+			modDVMTCalcFix.processKext(patcher, index, address, size);
```
vit9696, line 610: `processKext` is clearly wrong for the name of this function, as it only processes the framebuffer kext. At the very least I will ask it to be renamed to `processFramebufferKext`.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,199 @@
...
+		SYSLOG("igfx", "DVMT: This fix is not available on the current platform and has been disabled.");
+		return;
+	}
+		
+	// Enable the fix if designated boot argument or device property is found
+	enabled = checkKernelArgument("-igfxdvmt");
+	if (!enabled)
+		enabled = info->videoBuiltin->getProperty("enable-dvmt-calc-fix") != nullptr;
+	
+	// Read the DVMT preallocated memory set in BIOS from the GMCH Graphics Control field at 0x50 (PCI0,2,0)
```
vit9696, line 33: I do not believe this code should execute when `enabled` is false.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,199 @@
...
+	// TODO: Lilu needs to be updated to define the enumeration case `kIOPCIConfigGraphicsControl`
+	auto gms = WIOKit::readPCIConfigValue(info->videoBuiltin, /*WIOKit::kIOPCIConfigGraphicsControl*/ 0x50, 0, 16) >> 8;
```
vit9696, line 35: To read PCI configuration value you need call `WIOKit::awaitPublishing` as it will otherwise crash on Big Sur.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,199 @@
...
+		
+		// Guard: Should be able to disassemble the function
+		if (handle.flags & F_ERROR) {
+			SYSLOG("igfx", "DVMT: Failed to disassemble FBMemMgr_Init().");
+			break;
+		}
+		
+		// Instruction: shll $0x11, %???
+		// 3 bytes long if DSTReg < %r8d, otherwise 4 bytes long
+        if (handle.opcode == 0xC1 && handle.imm.imm8 == 0x11) {
```
vit9696, line 138: Something is up with the indentation here and below.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,199 @@
...
+		dvmt = gms * 32;
+	} else if (gms >= 0xF0 && gms <= 0xFE) {
+		dvmt = ((gms & 0x0F) + 1) * 4;
+	} else {
+		SYSLOG("igfx", "DVMT: GMS value is reserved. Check your BIOS settings. DVMT will be set to 0.");
+		return;
+	}
+	
+	DBGLOG("igfx", "DVMT: GMS value is 0x%02x; DVMT pre-allocated memory is %d MB.", gms, dvmt);
+	info->videoBuiltin->setProperty("fw-dvmt-gms-field-value", gms, 8);
```
vit9696, line 54: Is this for debugging? If so, guard it with ifdef DEBUG maybe?

WhateverGreen/kern_igfx.cpp
0xFireWolf: `processKext()` is defined in the interface `PatchSubmodule`. From the perspective of `PatchSubmodule`, it is possible that another concrete type implements this function to process the graphics accelerator kext instead.

Or do you suggest that we should modify the `PatchSubmodule` interface to separate kext processing into two functions, `processFramebufferKext()` and `processAcceleratorKext()`?

i.e. The new interface definition is as follows.

```
class PatchSubmodule {

    // ... Omitted ...

    virtual void processKernel(KernelPatcher &patcher, DeviceInfo *info) = 0;

    virtual void processFramebufferKext(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) = 0;

    virtual void processAcceleratorKext(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) = 0;
};
```

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf:
```
// Enable the fix if designated boot argument or device property is found
enabled = checkKernelArgument("-igfxdvmt");
if (!enabled)
	enabled = info->videoBuiltin->getProperty("enable-dvmt-calc-fix") != nullptr;
```

I believe that the code enables the patch if either the boot argument exists or the device property is defined.
I remember that this is the usual way to enable a patch, as I was corrected by someone else in a PR last year.
Are there any issues?

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: Not exactly for debugging, but I think it's better to make it only available in DEBUG version.

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: Fixed. Thanks for pointing that.

WhateverGreen/kern_igfx.cpp
vit9696: Probably the latter.

WhateverGreen/kern_igfx_memory.cpp
vit9696: I meant that the code after these checks has no real benefit to execute when `enable` is false.

WhateverGreen/kern_igfx.cpp
0xFireWolf: Done.

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: Ah, my bad. Sorry about that. I was thinking of always reading the DVMT value and adding it to the registry.

Fixed.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,212 @@
...
+			andlAddr = startAddress;
+			andlSize = handle.len;
+			andlDstr = (handle.rex_b << 3) | handle.modrm_rm;
+			SYSLOG("igfx", "DVMT: Found the andl instruction. Length = %d; DSTReg = %d.", andlSize, andlDstr);
+		}
+		
+		// Guard: Calculate and apply the binary patch if we have found both instructions
+		if (shllAddr && andlAddr) {
+			// Update the `movl` instruction with the actual amount of DVMT preallocated memory
+			*(uint32_t*) (movl + 2) = dvmt;
```
vit9696, line 168: Why does not this need enabling kernel writing? Is there anything wrong with the control flow?

Also, please use the C++ casts.

WhateverGreen/kern_igfx.hpp
```diff
@@ -499,6 +499,95 @@ class IGFX {
...
+		
+		/**
+		 *  True if this submodule requires patching the graphics acceleration driver
+		 */
+		bool requiresPatchingGraphics {false};
+		
+		/**
+		 *  Initialize any data structure required by this submodule if necessary
+		 */
+		virtual void init() = 0;
```
vit9696, line 531: Do not make (any) functions pure virtual in my opinion. Just provide empty body for them. This will make it much easier to implement in subsequent submodules.

WhateverGreen/kern_igfx.hpp
```diff
@@ -499,6 +499,95 @@ class IGFX {
...
+		/**
+		 *  Process the accelerator kext, retrieve and/or route functions if necessary
+		 *
+		 *  @param patcher KernelPatcher instance
+		 *  @param index   kinfo handle
+		 *  @param address kinfo load address
+		 *  @param size    kinfo memory size
+		 *  @note This funbction is called when the main IGFX module processes the kext.
+		 */
+		virtual void processAcceleratorKext(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) = 0;
```
vit9696, line 567: Is Accelerator kext — graphics kext? This is a bit not obvious as everything else uses graphics term. Maybe rename?

WhateverGreen/kern_igfx.cpp
```diff
@@ -593,6 +605,10 @@ bool IGFX::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t a
 			}
 		}
 
+		// We could iterate through each submodule and redirect the request if and only if the submodule is enabled
+		if (modDVMTCalcFix.enabled)
+			modDVMTCalcFix.processFramebufferKext(patcher, index, address, size);
```
vit9696, line 610: I believe a similar call to graphics kext is missing.

vit9696:
This looks good to me aside the last comments I left. I want to merge this into this release if possible.

WhateverGreen/kern_igfx.cpp
0xFireWolf: Yes, but not here. Fixed.

WhateverGreen/kern_igfx.hpp
0xFireWolf: Yes, it is the graphics kext. Sure, to make things consistent. 
Fixed.

WhateverGreen/kern_igfx.hpp
0xFireWolf: Make sense. A concrete type does not have to implement unnecessary functions.
Fixed.

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: `movl` is a local buffer that contains the bytes to be replaced. At this moment, we are not modifying kernel's memory. Later, we apply two patches via `lilu_os_memcpy()`.

Ah, I forgot to change the cast, as I was testing my patch in another C project. 
Fixed.

0xFireWolf:
> This looks good to me aside the last comments I left. I want to merge this into this release if possible.

Thanks for your reviewing. Let me know if there are any nuances to be fixed.

WhateverGreen/kern_igfx.hpp
```diff
@@ -499,6 +499,93 @@ class IGFX {
...
+	public:
+		/**
+		 *  Virtual destructor
+		 */
+		virtual ~PatchSubmodule() = default;
+		
+		/**
+		 *  True if this submodule should be enabled
+		 */
+		bool enabled {false};
```
07151129, line 516: Why duplicate this with `available`? Just don't set `enabled` if you want to disable the patch.

WhateverGreen/kern_igfx.hpp
```diff
@@ -499,6 +499,93 @@ class IGFX {
 		static bool forceWakeWaitAckFallback(uint32_t, uint32_t, uint32_t);
 		static void forceWake(void*, uint8_t set, uint32_t dom, uint32_t);
 	} ForceWakeWorkaround;
+	
+	/**
+	 *  Interface of a submodule to fix Intel graphics drivers
+	 */
+	class PatchSubmodule {
```
07151129, line 506: I'm unsure about this polymorphic design. It would be beneficial to store generalised patch pointers in a vector and iterate over those, but we are not doing that, so the advantage is lost.

WhateverGreen/kern_igfx.hpp
```diff
@@ -499,6 +499,93 @@ class IGFX {
...
+	public:
+		/**
+		 *  Virtual destructor
+		 */
+		virtual ~PatchSubmodule() = default;
```
07151129, line 511: When are we ever destructing a patch?

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,209 @@
...
+		patcher.clearError();
+		return;
+	}
+	
+	//
+	// Locate Apple's formula: (GGCRegValue << 0x11) & 0xFE000000
+	//
+	// macOS Catalina 10.15.6 (19G2021)
+	// loc_3f61f: movl $0x50, %esi
+	// loc_3f624: call __ZN11IOPCIDevice20extendedConfigRead16Ey
```
07151129, line 82: Can't we just use the hook of `extendedConfigRead` then, instead of patching the opcodes? WhateverGreen hooks  it anyway.

WhateverGreen/kern_igfx_memory.cpp
```diff
@@ -0,0 +1,209 @@
...
+			andlAddr = startAddress;
+			andlSize = handle.len;
+			andlDstr = (handle.rex_b << 3) | handle.modrm_rm;
+			SYSLOG("igfx", "DVMT: Found the andl instruction. Length = %d; DSTReg = %d.", andlSize, andlDstr);
+		}
+		
+		// Guard: Calculate and apply the binary patch if we have found both instructions
+		if (shllAddr && andlAddr) {
+			// Update the `movl` instruction with the actual amount of DVMT preallocated memory
+			*reinterpret_cast<uint32_t*>(movl + 2) = dvmt;
```
07151129, line 166: Is this preceded with `setKernelWriting`?

WhateverGreen/kern_igfx.hpp
0xFireWolf: My intention is that this fix is available for ICL but can be backported to older platforms. As far as I know, older laptops normally have DVMT set to 32MB or 64MB, but it is possible that a weird one sets it to other values which are not supported by the formula. I left this `available` property on purpose to prevent users from misusing the patch as well. (Not all users will carefully read the manual.)

WhateverGreen/kern_igfx.hpp
0xFireWolf: That's my proposal and expectation mentioned above. It would be easier to make new patches if we separate the giant IGFX module into multiple tiny submodules. Yes, ideally IGFX should store a collection of pointers to submodules and iterate over them.

WhateverGreen/kern_igfx.hpp
0xFireWolf: Previously, `PatchSubmodule` contains only pure virtual functions. As vit9696 suggests, I should make them non-pure and provide default empty implementations. I thought a virtual destructor is required if I want to make a class behave like an interface in C++. Maybe we no longer require the virtual destructor in this case?

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: No, we can't. The value read from PCI config space is valid but not compatible with Apple's formula. The value indicates the actual amount of DVMT memory preallocated by BIOS (e.g. 60MB), and we cannot fake it to pretend that we have allocated 64MB instead.

WhateverGreen/kern_igfx_memory.cpp
0xFireWolf: See above: https://github.com/acidanthera/WhateverGreen/pull/64#discussion_r480917666.

WhateverGreen/kern_igfx.hpp
07151129: What I meant is that patch deallocation newer appears to happen in WhateverGreen anyway.

WhateverGreen/kern_igfx.hpp
07151129: So why can't you use the single variable, `enabled`, for exactly that purpose?

vit9696: Because they are split in time?

0xFireWolf: Because `enabled` can be controlled by the boot argument or the device property, and then it is hard to know whether an user is trying to enable the patch on an unavailable platform.

WhateverGreen/kern_igfx_memory.cpp
07151129: What if you report a lower multiple of 32MB instead?

0xFireWolf: Based on my experiment, 32MB is too small for some laptops, such as Surface Pro 7, with a high resolution display. There must be a reason why OEM chooses 60MB as a default value for DVMT preallocated memory on Ice Lake, so I think it's better to use their default value instead of trying to lower it down. (Well, if one could modify their BIOS settings via EFI shell or RU.EFI, they could always choose a higher value.)

07151129: I see. Perhaps this could be implemented in OpenCore then? That would simplify the kernelspace part quite a bit...

vit9696: It cannot be.

07151129: To document this a little, the problem is that DVMT config via the GMS register is impossible after the register gets locked for read-only access in PEI. Thus neither the bootloader nor the kernel can reconfigure it, and we must resort to teaching the driver to work with the given size. It may still be possible to provide a different size using OpRegion, but I wouldn't rely on that.

kocoman1:
A bit offopic but I don't know how to contact FireWolf, I hope vit9696 will keep this thread 
I have a HD4000 Ivy Bridge that installed coreboot, I get this panic.

AppleIntelCapriController::FBMemMgr_Init(AppleIntelCapriController *this)

/Library/Caches/com.apple.xbs/Sources/GPUDriversIntel/GPUDriversIntel-14.5.22/IONDRV/IVB/AppleIntelFramebuff"
            "er/AppleIntelController.cpp:14147 Assertion failed: (0 == i) || ((pfbInfo->rangeStolen.physBase - pfbInfoPre"
            "v->rangeStolen.physBase) == fFramebufferMemorySize) \

I don't understand what the error above means, and also from the decompiled AppleIntelFramebufferCapri.c code from ida, but can anyone tell me whats wrong, I ran the -igfxdvmt (after some patch to force it to run) and it says my Dvmt is 64mb and no patch needed.

Is it from this?

(linux with 0x06 debug)
[    8.994523] [drm:i915_gem_object_create_stolen_for_preallocated [i915]] creating preallocated stolen object: stolen_offset=0x0000000000000000, gtt_offset=0x0000000000000000, size=0x0000000000500000

[    9.005900] [drm:i915_init_ggtt [i915]] clearing unused GTT space: [501000, 80000000]

vs


[    9.558873] [drm:i915_gem_object_create_stolen_for_preallocated [i915]] creating preallocated stolen object: stolen_offset=0x0000000000000000, gtt_offset=0x0000000000000000, size=0x0000000000300000

[    9.575057] [drm:i915_init_ggtt [i915]] clearing unused GTT space: [301000, 80000000]

The size difference 500000 vs 300000?

I can try to nop the jz so it skips the debugger call and see what happens..

Thanks for any hints

WhateverGreen/kern_igfx_memory.cpp
1alessandro1: @vit9696 Could this be implemented as a separate driver loaded from the "ConnectDrivers" section on OpenCore's config.plist?

1alessandro1: Instead of implementing it in OpenCore.efi main driver

CobanRamo:
Hello @0xFireWolf ,
I hope that I'm right here
I think with this build (https://github.com/acidanthera/WhateverGreen/commit/80f4727765d79e7ce0c0d2dac4e2517981e50138) there are new problems with i7-1065G7 Graphic.
With this build, Catalina 15.7 & BigSur Beta 10 will no longer start.
With the previous build I was able to install and start both OS without problems.
With CT the black screen remains, with BS there is a panic restart.

m0d16l14n1:
@CobanRamo remove boot-args of properties for LSPCON and Enhanced Maximum Link Rate Fix

I got black screen too, if there are boot-args to enable that fixes

eslamm2500:
first i got painc and the last update fixed it with 2 boot-arg and it load
succes but many times open in black screen so now what the fix ?
intel 1065g7

CobanRamo:
@m0d16l14n1 

actually, should have figured it out myself.
Can confirm, without boot-args or properties everything works again.

Thank you

eslamm2500:
some people told me that thay face the same problem (black screen) and
fixed by old version af lilu or vituralsmc
the problem is the laptop open and dedect the graphic card but when it open
there is some thing like glitch and when restart it gives me black screen
many times (like it power off not low brightness ) (tried it without device
brightness and got the same problem )
the laptob is dell 3593 with 1065g7

PR #66, opened 2020-09-06
Update FAQ.GeForce.en.md

ashishpatel1992:
Not mentioned in which file we need to add Device (PEG0){ .....


ashishpatel1992:
@vit9696 can you mention the filename in which we need to the above configuration?

vit9696:
There is no particular filename, it is just ACPI.

PR #69, opened 2020-09-19
New way to patch backlight on Kaby Lake and older

hieplpvip:
Hi, this pull request adds a new backlight patch for Kaby Lake and older based on the Coffee Lake backlight patch. It removes the need of injecting custom PWMMax with SSDT-PNLF, allowing using the default PWMMax system firmware.

With this patch, we can use SSDT-PNLFCFL for any platforms, just need to change _UID.

Tested on KabyLake (HD620)

P/s: This is my first time working with Intel graphics patching. If I did anything wrong, please help me correct it. Thank you very much!

### DIFF
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -328,6 +328,15 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
 		else if (currentFramebuffer == &kextIntelCFLFb && BaseDeviceInfo::get().modelType == WIOKit::ComputerModel::ComputerLaptop)
 			cflBacklightPatch = CoffeeBacklightPatch::Auto;
 
+		// Enable new backlight patch type on KBL and older if the corresponding boot argument is found
+		newBacklightPatch = checkKernelArgument("-igfxnewbklt");
+		// Or if "enable-new-backlight-fix" is set in IGPU property
+		if (!newBacklightPatch)
+			newBacklightPatch = info->videoBuiltin->getProperty("enable-new-backlight-fix") != nullptr;
+		// New backlight patch is only for KBL and older
+		if (newBacklightPatch && cpuGeneration > CPUInfo::CpuGeneration::KabyLake)
+			SYSLOG("igfx", "new backlight patch is only for KBL and older. CFL and above must use another patch.");
+
 		if (WIOKit::getOSDataValue(info->videoBuiltin, "max-backlight-freq", targetBacklightFrequency))
 			DBGLOG("igfx", "read custom backlight frequency %u", targetBacklightFrequency);
 
@@ -380,6 +389,8 @@ void IGFX::processKernel(KernelPatcher &patcher, DeviceInfo *info) {
 			if (dumpFramebufferToDisk || dumpPlatformTable || debugFramebuffer)
 				return true;
 			if (cflBacklightPatch != CoffeeBacklightPatch::Off)
+				return true;
+			if (newBacklightPatch)
 				return true;
 			if (maxLinkRatePatch)
 				return true;
@@ -502,14 +513,14 @@ bool IGFX::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t a
 		// Accept Kaby FB and enable backlight patches if On (Auto is irrelevant here).
 		bool bklKabyFb = realFramebuffer == &kextIntelKBLFb && cflBacklightPatch == CoffeeBacklightPatch::On;
 		// Solve ReadRegister32 just once as it is shared
-		if (bklCoffeeFb || bklKabyFb ||
+		if (bklCoffeeFb || bklKabyFb || newBacklightPatch ||
 			RPSControl.enabled || ForceWakeWorkaround.enabled || coreDisplayClockPatch) {
 			AppleIntelFramebufferController__ReadRegister32 = patcher.solveSymbol<decltype(AppleIntelFramebufferController__ReadRegister32)>
 			(index, "__ZN31AppleIntelFramebufferController14ReadRegister32Em", address, size);
 			if (!AppleIntelFramebufferController__ReadRegister32)
 				SYSLOG("igfx", "Failed to find ReadRegister32");
 		}
-		if (bklCoffeeFb || bklKabyFb ||
+		if (bklCoffeeFb || bklKabyFb || newBacklightPatch ||
 			RPSControl.enabled || ForceWakeWorkaround.enabled) {
 			AppleIntelFramebufferController__WriteRegister32 = patcher.solveSymbol<decltype(AppleIntelFramebufferController__WriteRegister32)>
 			(index, "__ZN31AppleIntelFramebufferController15WriteRegister32Emj", address, size);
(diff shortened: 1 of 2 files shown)


hieplpvip:
Sample `SSDT-PNLF`:
```asl
// Adding PNLF device for WhateverGreen.kext and others.
// Rename GFX0 to anything else if your IGPU name is different.

// Change _UID to match profiles in WhateverGreen.kext Info.plist
// 14: Sandy/Ivy
// 15: Haswell/Broadwell
// 16: Skylake/KabyLake (and some Haswell, example 0xa2e0008)
// 17: custom LMAX=0x7a1
// 18: custom LMAX=0x1499
// 19: CoffeeLake
// 99: Other (requires custom AppleBacklightInjector.kext/WhateverGreen.kext)

DefinitionBlock ("", "SSDT", 2, "ACDT", "PNLF", 0x00000000)
{
    External (_SB_.PCI0.GFX0, DeviceObj)

    Device (_SB.PCI0.GFX0.PNLF)
    {
        Name (_ADR, Zero)  // _ADR: Address
        Name (_HID, EisaId ("APP0002"))  // _HID: Hardware ID
        Name (_CID, "backlight")  // _CID: Compatible ID
        Name (_UID, 0x10)  // _UID: Unique ID
        Name (_STA, 0x0B)  // _STA: Status
    }
}
```

07151129:
I may be missing something, but what's the advantage of this method if it requires injecting custom SSDT anyway?

Andrey1970AppleLife:
[SSDT-PNLF](https://github.com/acidanthera/WhateverGreen/blob/master/Manual/SSDT-PNLF.dsl) injects _UID and PWMMAX automatically depending on a platform, аnd in most cases doesn't require editing.

hieplpvip:
This patch allows display backlight to work right after boot on LG gram 13Z970. With traditional `SSDT-PNLF`, it only works after a sleep/wake cycle. I can't find what's wrong with `SSDT-PNLF` (perhaps some registers are initialized incorrectly), so I created this patch. I think this is a cleaner way to patch backlight.

goodbest:
> This patch allows display backlight to work right after boot on LG gram 13Z970. With traditional `SSDT-PNLF`, it only works after a sleep/wake cycle. I can't find what's wrong with `SSDT-PNLF` (perhaps some registers are initialized incorrectly), so I created this patch. I think this is a cleaner way to patch backlight.

My dell 7380 laptop (7200u, 1080p screen, HD620, hardware almost the same with your LG machine.) used to have the same issue.
After a lot of work, I figured out that I have to use [`SSDT-RMCF`](https://github.com/RehabMan/OS-X-Clover-Laptop-Config/blob/master/hotpatch/SSDT-RMCF.dsl) together with `SSDT-PNLF` to solve this issue, which I think `perhaps some registers are initialized incorrectly` might be the cause.

In my `SSDT-RMCF`, I changed the variables to the following values. You can also try different combinations for that.
```
Name (HIGH, One)
Name (BKLT, 0x05)
Name (LMAX, 0x056C)
```

vit9696:
Well, the current `SSDT-PNLF` is more convenient for one reason: it detects the platform and sets the correct `_UID` automatically. If anything, the new SSDT should preserve this functionality. I would also prefer to unify `SSDT-PNLF` with `SSDT-PNLFCFL` to have one SSDT for all since we are changing this anyway.

As for the approach itself, from what I understand the problem is that macOS drivers hardcodes the PWM frequency and thus the range of the backlight level. There are two ways to avoid it:

- Set the PWM frequency to the one macOS expects (this is what SSDT-PNLF does).
- Rescale the values used by macOS to the frequency set by the BIOS GOP code (this is what the CFL code does).

The second approach might be better as it will avoid the backlight flashes during the startup, but I wonder what downsides does it actually have besides the WEG patches. Could you provide some more details on the implementation and the problem as we do not know much about it.

goodbest:
> Well, the current `SSDT-PNLF` is more convenient for one reason: it detects the platform and sets the correct `_UID` automatically. If anything, the new SSDT should preserve this functionality. I would also prefer to unify `SSDT-PNLF` with `SSDT-PNLFCFL` to have one SSDT for all since we are changing this anyway.
> 
> As for the approach itself, from what I understand the problem is that macOS drivers hardcodes the PWM frequency and thus the range of the backlight level. There are two ways to avoid it:
> 
> * Set the PWM frequency to the one macOS expects (this is what SSDT-PNLF does).
> * Rescale the values used by macOS to the frequency set by the BIOS GOP code (this is what the CFL code does).
> 
> The second approach might be better as it will avoid the backlight flashes during the startup, but I wonder what downsides does it actually have besides the WEG patches. Could you provide some more details on the implementation and the problem as we do not know much about it.

I agree that using ``SSDT-PNLF` would be more convenient.
However I encountered the same issue with @hieplpvip I described above, so I suppose there must be some bug (or missing variable that should be added) in current PNLF DSL file.

1alessandro1:
Sorry for the little ot, but SSDT-PLNF and SSDT-PLNFCFL apply only to laptops or for desktops too?
I've heard the only way to control desktop brightness on desktops was to check the monitor itself (its internal parameters)... If this is true and SSDT-PLNF/SSDT-PNLFCFL are not required, does macOS have brightness settings at 100% all the time (leaning the brighness controls to the monitor panel only)?

hieplpvip:
> Well, the current `SSDT-PNLF` is more convenient for one reason: it detects the platform and sets the correct `_UID` automatically. If anything, the new SSDT should preserve this functionality. I would also prefer to unify `SSDT-PNLF` with `SSDT-PNLFCFL` to have one SSDT for all since we are changing this anyway.

This should be trivial. 

> As for the approach itself, from what I understand the problem is that macOS drivers hardcodes the PWM frequency and thus the range of the backlight level. There are two ways to avoid it:
> 
> * Set the PWM frequency to the one macOS expects (this is what SSDT-PNLF does).
> * Rescale the values used by macOS to the frequency set by the BIOS GOP code (this is what the CFL code does).
> 
> The second approach might be better as it will avoid the backlight flashes during the startup, but I wonder what downsides does it actually have besides the WEG patches. Could you provide some more details on the implementation and the problem as we do not know much about it.

My patch uses the second way. I just copied the CFL code and changed the registers to match the platforms.
```
Ivy Bridge and before use two registers: BXT_BLC_PWM_FREQ1 for frequency and BLC_PWM_CPU_CTL for duty cycle
Haswell to Kaby Lake use only one register: BXT_BLC_PWM_FREQ1 for both frequency and duty cycle
```

MrHomebrew:
I could definitely test to see if this is working. Currently I have the surface pro 2017 with Intel I7 hd620 and with OpenCire I keep getting black screen on internal. This seems to be a possible solution

MrHomebrew:
No matter what I do I can not get the internal display to fully light up. Black screen with backlight it seems. Idk what is going on and being a hidpi display the resolutions that become stock are all for 1080p monitors. My internal in 2736x1824.

Idk what to do

hieplpvip:
@MrHomebrew Did you enable the patch (with bootarg `-igfxnewbklt` or device property `enable-new-backlight-fix`)?

MrHomebrew:
I had no idea about that boot argument or device property. I will definitely give it a try. Where did you find that info?

hieplpvip:
> I had no idea about that boot argument or device property. I will definitely give it a try. Where did you find that info?

It's in the code...

hieplpvip:
I will close this as I'm working on a dedicated kext for backlight [here](https://github.com/hieplpvip/AppleBacklightSmoother).

vit9696:
Ugh, why do we need a billion of kexts that duplicate the functionality?

hieplpvip:
I'm trying to get smooth backlight transition working. I couldn't figure out how to let macOS do it natively so I have to do it manually by splitting the big change in backlight duty cycle into consecutive small enough changes (e.g. `978 -> 1028 -> 1079 -> 1132` instead of `978 -> 1132`). There is a 10 millisecond delay between each change.

It's working pretty well for me, though I think I can fine tune it more. The code is kinda hacky so I think it should be a dedicated plugin for now instead of being a part of WhateverGreen.

MrHomebrew:
hieplpvip  cou)d you take a look at my OpenCore EFI folder and possibly make the necessary changes this way I can reflect and understand what I did wrong that I can correct in the future.

Here is a link below

https://github.com/MrHomebrew/-Surface-Pro-2017-OpenCore-EFI-

If anyone can help get this internal display to turn on you will really be a big help.

1alessandro1:
@hieplpvip I suggest to adapt the AppleBacklightSmoother code to WhateverGreen as @vit9696 suggested, since having two kexts handling AppleIntelxxxGraphics modifications is not ideal

PR #71, opened 2020-10-13
fiX ssdt pnlf to include cfl

Lorys89:


Lorys89:
I am attaching aml version

[SSDT-PNLF-ALL.zip](https://github.com/acidanthera/WhateverGreen/files/5373650/SSDT-PNLF-ALL.zip)

Lorys89:
WHY?

Andrey1970AppleLife:
Use [SSDT-PNLFCFL](https://github.com/acidanthera/WhateverGreen/blob/master/Manual/SSDT-PNLFCFL.dsl)

Lorys89:
> Use [SSDT-PNLFCFL](https://github.com/acidanthera/WhateverGreen/blob/master/Manual/SSDT-PNLFCFL.dsl)

but ssdt pnlf all works, even on cfl notebook

vit9696:
Are we sure it does not break other laptops? I believe we did not update PNLF, because CFL additions caused issues to older models.

Lorys89:
tried on notebook cfl and kbl, now text on haswell notebook and I'll let you know.

vit9696:
CC @hieplpvip @headkaze @zhen-zen @khronokernel what can you add? I do not remember much about the backlight. Will reopen for now till I get further comment.

Lorys89:
tried on haswell notebooks and works.

zhen-zen:
Could you briefly describe where is the new `COFFEELAKE_PWMMAX` from? Other parts seems good since it will only affect CFL. So there's no need for testing on other platforms. And please fix the indent in `_STA`. 

Also, I just found something interesting: [Dynamic LED Backlight Control and Debug on Embedded Display Port (eDP) Panels using PWM Control Techniques](https://cdrdv2.intel.com/v1/dl/getContent/335207). Maybe the doc could help with the explanation, or if it's possible to make it part of the driver? Although I have written some code to reduce ACPI patches, current `SSDT-PNLF` is not that invasive and does a good job.

Lorys89:
> Could you briefly describe where is the new `COFFEELAKE_PWMMAX` from? Since the modification will only affect CFL, there's no need for testing on other platforms. And please fix the indent in `_STA`.
> 
> Also, I just found something interesting: [Dynamic LED Backlight Control and Debug on Embedded Display Port (eDP) Panels using PWM Control Techniques](https://cdrdv2.intel.com/v1/dl/getContent/335207). Maybe the doc could help with the explanation, or if it's possible to make it part of the driver? Although I have written some code to reduce ACPI patches, current `SSDT-PNLF` is not that invasive and does a good job.

the value is taken from cfl notebook, tested and works ssdt, ffff was not ok, ff3e is ok, if you need anything else, just ask.

Lorys89:
@zhen-zen I await your instructions, I am of the opinion that there must be only one universal ssdt pnlf

Manual/SSDT-PNLF.dsl
```diff
@@ -235,7 +245,7 @@ DefinitionBlock("", "SSDT", 2, "ACDT", "PNLF", 0)
                 If (!(8 & Local4) && Local2 != Local1)
                 {
                     // set new backlight PWMAX but retain current backlight level by scaling
-                    Local0 = (((^LEVX & 0xFFFF) * Local2) / Local1) | (Local2 << 16)
+                    Local0 = (((^LEVX & 0xFF3E) * Local2) / Local1) | (Local2 << 16)
```
hieplpvip, line 248: This change is wrong. This line of code is for Haswell to Kaby Lake. On these platforms, both backlight max frequency and duty cycle are controlled using only one 32-bit register: low 16 bits are for duty cycle and high 16 bits are for max frequency.

`^LEVX & 0xFFFF` gives us the low 16 bits, or duty cycle. We then rescale this value to match the new PWMMAX being set.

hieplpvip:
The traditional way to patch backlight before WhateverGreen exist is to set the hardware PWMMAX to match the value specified in the framebuffer (e.g. 0x56c for Kaby Lake). This is done with SSDT-PNLF.

For some reasons (I can't remember exactly, seem to be 32-bit calculation overflow), this doesn't work anymore on Coffee Lake. Instead of setting the hardware PWMMAX to match that of macOS, we now rescale the value from macOS to match the hardware ([source](https://github.com/acidanthera/WhateverGreen/blob/master/WhateverGreen/kern_igfx.cpp#L522-L570)). This allows SSDT-PNLFCFL to be much cleaner, as we only need to set _UID to match backlight profiles [here](https://github.com/acidanthera/WhateverGreen/blob/master/WhateverGreen/kern_weg.cpp#L32-L89).

I don't know where the value `COFFEELAKE_PWMMAX 0xffff` comes from. According to the comments in WhateverGreen, macOS uses 7777 or 22222 for Coffee Lake. I guess 0xffff is some arbitrary value that RehabMan chose. Still, neither `0xffff` nor `0xff3e` match the value used by macOS. This may result in not having full range of backlight.

I believe that the new patch should also be ported back to older platforms. @vit9696 I can do that if you want.

vit9696:
@hieplpvip if you mean like your previous PR, that makes sense to me. Last time we did not test the changes in time and compare them to the original, but that is what we wanted to do before merging. I agree that this CFL approach is cleaner than what we have now, so better to backport it I guess.

Closing this PR due to issues explained above. For _STA change this could be submitted separately.

Andrey1970AppleLife:
> For _STA change this could be submitted separately.

It isn't mandatory for this table.

vit9696:
Does not Windows show unidentified device without it? If not, I guess we can pass indeed.

Andrey1970AppleLife:
> Does not Windows show unidentified device without it? If not, I guess we can pass indeed.

No problems with Windows.

Lorys89:
> Does not Windows show unidentified device without it? If not, I guess we can pass indeed.

revert _sta  https://github.com/Lorys89/WhateverGreen/blob/master/Manual/SSDT-PNLF.dsl

PR #72, opened 2020-10-17
Enhanced maximum link rate fix, Fixed the page fault in the LSPCON driver, Submodules refactoring.

0xFireWolf:
**This pull request contains 2 new features, 1 bug fix, and 5 code refactoring changes on submodules.**

### >> Enhanced Maximum Link Rate Fix

**Abstract:**

Users are no longer required to specify a maximum link rate for their builtin display via the `dpcd-max-link-rate` property. The "invalid" maximum link rate reported by the display panel has a special meaning in the eDP 1.4 standard. eDP 1.4 panels may support more link rates other than traditional ones such as RBR, HBR, HBR2 and HBR3. As such, a table of all supported rates can be found in another DPCD register. This update adds a new probe function to find and check the maximum link rate for eDP 1.4 panels, and hence eliminates the need of a manual value injection. The previous solution is used as a fallback mechanism, in case the probed value is not supported by the graphics driver. Additionally, this fix is extended to support Ice Lake platform.

**Related Research Article:** 

[Coffee Lake Intel UHD Graphics 630 on macOS Catalina: The ultimate solution to the kernel panic due to division by zero in the framebuffer driver](https://www.firewolf.science/2020/10/coffee-lake-intel-uhd-graphics-630-on-macos-catalina-the-ultimate-solution-to-the-kernel-panic-due-to-division-by-zero-in-the-framebuffer-driver/)

### >> Fixed Page Fault in the LSPCON Driver

**Abstract:**

`ReadAUX()` is resolved for the maximum link rate fix but not for the LSPCON driver. As a result, if the maximum link rate fix is disabled, the function pointer is `NULL` and hence a page fault occurs. LSPCON driver needs `ReadAUX()` to wake up the DisplayPort AUX channel for the adapter. This issue is addressed by resolving the symbol for the LSPCON driver as well.

### >> Submodules Refactoring

**Abstract:**

- The maximum link rate fix, core display clock fix and HDMI dividers calculation fix are now implemented as patch submodules. A new file `kern_igfx_clock.cpp` is created to implement these clock-related fixes.
- The I2C-over-AUX transaction APIs are separated from the LSPCON driver into a standalone submodule. Now one can enable verbose output in I2C transactions without enabling the LSPCON driver. It is also convenient to develop new patches that rely on I2C APIs.
- The LSPCON driver has been relocated to a standalone source unit `kern_igfx_lspcon.cpp/hpp`.

In general, the size of IGFX is greatly reduced, and it is more clear to identify which functions are routed by a submodule.


WhateverGreen/kern_igfx.hpp
```diff
@@ -570,1062 +508,703 @@ class IGFX {
...
-		 *  @param mode The expected mode; one of `LS` and `PCON`
-		 */
-		inline bool isRunningInMode(Mode mode) {
-			Mode currentMode;
-			if (getMode(&currentMode) != kIOReturnSuccess) {
-				DBGLOG("igfx", "LSPCON::isRunningInMode() Error: [FB%d] Failed to get the current adapter mode.", index);
-				return false;
-			}
-			return mode == currentMode;
+		inline bool hasLSPCON(uint32_t index) {
```
vit9696, line 999: When declaring a class function within the class you can safely omit `inline`, this is implied.

WhateverGreen/kern_igfx_clock.cpp
```diff
@@ -0,0 +1,881 @@
...
+	DBGLOG("igfx", "MLR: [COMM] wrapReadAUX() Maximum link rate 0x%02x has been set in the DPCD buffer.", caps->maxLinkRate);
+	return retVal;
+}
+
+IOReturn IGFX::DPCDMaxLinkRateFix::orgReadAUX(uint32_t address, void *buffer, uint32_t length) {
+	if (port != nullptr) {
+		// ICL+
+		DBGLOG("igfx", "MLR: [COMM] orgReadAUX() Routed to ICL IMP with Address = 0x%x; Length = %u.", address, length);
+		return orgICLReadAUX(port, address, buffer, length);
+	} else {
```
vit9696, line 267: Better drop the else, the upper if returns anyway.

WhateverGreen/kern_igfx_clock.cpp
```diff
@@ -0,0 +1,881 @@
...
+		// CFL-
+		DBGLOG("igfx", "MLR: [COMM] orgReadAUX() Routed to CFL IMP with Address = 0x%x; Length = %u.", address, length);
+		return orgCFLReadAUX(controller, framebuffer, address, length, buffer, displayPath);
+	}
+}
+
+bool IGFX::DPCDMaxLinkRateFix::getFramebufferIndex(uint32_t &index) {
+	auto framebuffer = port != nullptr ? orgICLGetFBFromPort(*callbackIGFX->gFramebufferController, port) : this->framebuffer;
```
vit9696, line 275: Ugh, better not to alias class members. Better make it fb or something like that for the local variable.

WhateverGreen/kern_igfx_clock.cpp
```diff
@@ -0,0 +1,881 @@
...
+	uint16_t rates[DP_MAX_NUM_SUPPORTED_RATES] = {0};
+	if (orgReadAUX(DPCD_EDP_SUPPORTED_LINK_RATES_ADDRESS, rates, sizeof(rates)) != kIOReturnSuccess) {
+		SYSLOG("igfx", "MLR: [COMM] ProbeMaxLinkRate() Failed to read supported link rates from DPCD.");
+		return 0;
+	}
+	
+	// Parse all supported link rates reported by DPCD
+	// The last non-zero entry in the table is the maximum link rate supported by the eDP 1.4 panel
+	uint32_t last = 0;
+	for (int index = 0; index < arrsize(rates); index += 1) {
```
vit9696, line 306: Maybe `index++`?

More importantly, is it guaranteed that the table is sorted? In the description you said we choose the highest value.

WhateverGreen/kern_igfx_i2c_aux.cpp
```diff
@@ -0,0 +1,187 @@
...
+	
+	// Or if verbose output in transactions is enabled
+	if (checkKernelArgument("-igfxi2cdbg")) {
+		enabled = true;
+		verbose = true;
+	}
+}
+
+void IGFX::AdvancedI2COverAUXSupport::processFramebufferKext(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) {
+	auto roa = patcher.solveSymbol(index, "__ZN31AppleIntelFramebufferController14ReadI2COverAUXEP21AppleIntelFramebufferP21AppleIntelDisplayPathjtPhbh", address, size);
```
vit9696, line 30: Will not it be easier/more readable to use routeMultiple here and in similar places above?

WhateverGreen/kern_igfx.cpp
```diff
@@ -503,7 +437,7 @@ bool IGFX::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t a
 		bool bklKabyFb = realFramebuffer == &kextIntelKBLFb && cflBacklightPatch == CoffeeBacklightPatch::On;
 		// Solve ReadRegister32 just once as it is shared
 		if (bklCoffeeFb || bklKabyFb ||
-			RPSControl.enabled || ForceWakeWorkaround.enabled || coreDisplayClockPatch) {
+			RPSControl.enabled || ForceWakeWorkaround.enabled || modCoreDisplayClockFix.enabled) {
```
vit9696, line 440: This is a bit ugly. Probably we want a separate bool for submodules like `requiresGenericRegisterIo`. We can make a joint version of this bool in IGFX, assign it in processKernel as an OR of all the submodules values, and then use here.

 For now since not all entries are submodules we can assign it some hardcoded entries too. But please add a FIXME.

WhateverGreen/kern_igfx.hpp
```diff
@@ -570,1062 +508,703 @@ class IGFX {
...
+		void init() override;
+		void deinit() override;
+		void processKernel(KernelPatcher &patcher, DeviceInfo *info) override;
+		void processFramebufferKext(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) override;
+	} modLSPCONDriverSupport;
+	
+	/**
+	 *  The LSPCON driver requires access to I2C-over-AUX transaction APIs
+	 */
+	friend class LSPCON;
```
vit9696, line 1066: Cannot private classes in C++(ones declared inside another) access private fields of their owner? Maybe I am forgetting something.

vit9696 (commented):
Thanks for the patches, particularly the submodules. Makes it better as we go.

The changes look good. The refactored code was mostly there, so I did not review the it too precisely. Noted a few issues, mostly stylistic.

Also, could you do a history rebase next time, like git rebase origin/master? Would be easier to handle the changes with cleaner commit history.

WhateverGreen/kern_igfx.hpp
```diff
@@ -505,11 +448,6 @@ class IGFX {
 	 */
 	class PatchSubmodule {
 	public:
-		/**
-		 *  Virtual destructor
```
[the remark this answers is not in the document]
0xFireWolf: In my previous pull request, @07151129 indicated that we would never destruct a submodule https://github.com/acidanthera/WhateverGreen/pull/64#discussion_r481390611. I understand that virtual destructor is required if we `delete` an instance of its derived type, so his suggestion convinced me.

However, my original intention is that we should always declare a virtual destructor if we make a class behave as an interface, even though we never construct an instance of its derived type dynamically. I have added it back in my latest commit. What's your suggestion or opinion?

WhateverGreen/kern_igfx_clock.cpp
0xFireWolf: Based on my observation so far, the table is sorted in ascending order, so the last non-zero element should be the maximum link rate supported by the panel.

I have added a guard to ensure that we will get the largest value from the table, in case that a "hateful" panel does not conform to the standard.

WhateverGreen/kern_igfx.hpp
0xFireWolf: `LSPCON` is a standalone class declared in `kern_igfx_lspcon.hpp` and is no longer nested in `IGFX`.

WhateverGreen/kern_igfx_clock.cpp
```diff
@@ -0,0 +1,879 @@
...
+	orgIclReadRegister32 = reinterpret_cast<decltype(orgIclReadRegister32)>(callbackIGFX->AppleIntelFramebufferController__ReadRegister32);
+	
+	if (patcher.routeMultiple(index, &request, 1, address, size) && orgDisableCDClock && orgSetCDClockFrequency && orgIclReadRegister32) {
+		DBGLOG("igfx", "CDC: Functions have been routed successfully.");
+	} else {
+		patcher.clearError();
+		SYSLOG("igfx", "CDC: Failed to route functions.");
+	}
+}
+
```
0xFireWolf, line 503: @vit9696 This is a follow-up to your review https://github.com/acidanthera/WhateverGreen/pull/72#discussion_r506909740.

I have updated my code to use `routeMultiple()` whenever possible, and now those functions are more readable.

However, I don't think the patcher supports solving multiple symbols in one shot. Take the above function as an example, the fix needs to resolve two more functions `disableCDClock()` and `setCDClockFreq()`. 

I would like to propose a new Lilu API `solveMultiple()` that is similar to `routeMultiple()`. The new API takes an array of `SolveRequest` and resolve all symbols in one function call. As a result, it would greatly simplify the case where multiple functions are required as helpers to perform a fix. 

What's your opinion?

WhateverGreen/kern_igfx.cpp
0xFireWolf: Totally agreed. At this moment, backlight fix is the only one that injects code to `readRegister32/writeRegister32`, while others only need them as helper functions. 

The worst case is that multiple fixes request to inject code to those functions, so we probably need something like `GenericRegisterModifier` to orchestrate injection requests. The same thing might happen in the AUX register case.

For the global controller access, we could apply the same concept as well. Apple has refactored a lot of code from `AppleIntelFramebufferController` to the new class `AppleIntelPort`, so a new fix might need the controller instance to fetch the framebuffer instance associated with a port via `AppleIntelFramebufferController::getFBFromPort()`.

WhateverGreen/kern_igfx.hpp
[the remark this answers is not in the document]
vit9696: Yes, it is safer to have it, because somebody might later add a destruction and forget to add a virtual destructor.

WhateverGreen/kern_igfx.hpp
vit9696: Ah, fine then.

WhateverGreen/kern_igfx_clock.cpp
vit9696: Hmmm, I do not mind, but I may not be able to work on it for the next week at least. If you send the patches we can make them land though.

0xFireWolf: OK, I will do that and submit a PR to Lilu later.

0xFireWolf: I have submitted a PR (https://github.com/acidanthera/Lilu/pull/61) to Lilu, and I also updated some code to use the new API in WhateverGreen. Finally, everything is clean and no more annoying type castings.

0xFireWolf:
@vit9696 I think that's all for this PR. I will do a history rebase next time, and I will take a look at the register refactoring (https://github.com/acidanthera/WhateverGreen/pull/72#discussion_r506913581) once I have time. The current RPS control fix and the force wake workaround are already nicely packed, so I think it would be easy to port them as a submodule. 

Let me know if there any further changes I need to make before this PR is merged.

Thanks for your time.

vit9696:
routeMultiple and solveMultiple should not leave errors after them, so you do not need clearError. The rest looks good.

0xFireWolf:
> routeMultiple and solveMultiple should not leave errors after them, so you do not need clearError. The rest looks good.

Done.

PR #75, opened 2020-11-25
Docs: Only Darwin for PNLF

Lorys89:
add _sta method return for win acpi

### DIFF
--- a/Manual/SSDT-PNLF.dsl
+++ b/Manual/SSDT-PNLF.dsl
@@ -46,7 +46,17 @@ DefinitionBlock("", "SSDT", 2, "ACDT", "PNLF", 0)
         // 19: CoffeeLake 0xffff
         // 99: Other (requires custom AppleBacklightInjector.kext/WhateverGreen.kext)
         Name(_UID, 0)
-        Name(_STA, 0x0B)
+        Method (_STA, 0, NotSerialized)
+            {
+                If (_OSI ("Darwin"))
+                {
+                    Return (0x0B)
+                }
+                Else
+                {
+                    Return (Zero)
+                }
+            }
 
         Field(^RMP3, AnyAcc, NoLock, Preserve)
         {
--- a/Manual/SSDT-PNLF.dsl
+++ b/Manual/SSDT-PNLF.dsl
@@ -46,17 +46,17 @@ DefinitionBlock("", "SSDT", 2, "ACDT", "PNLF", 0)
         // 19: CoffeeLake 0xffff
         // 99: Other (requires custom AppleBacklightInjector.kext/WhateverGreen.kext)
         Name(_UID, 0)
-        Method (_STA, 0, NotSerialized)
+        Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
+            If (_OSI ("Darwin"))
             {
-                If (_OSI ("Darwin"))
-                {
-                    Return (0x0B)
-                }
-                Else
-                {
-                    Return (Zero)
-                }
+                Return (0x0B)
             }
+            Else
+            {
+                Return (Zero)
+            }
+        }
 
         Field(^RMP3, AnyAcc, NoLock, Preserve)
         {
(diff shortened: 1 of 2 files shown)


Lorys89:
IMG_20201126_003434

Lorys89:
putting method _sta with darwin, on windows there is no more error.

1alessandro1:
I still don't get why you use OpenCore to boot Windows. That's just dumb.

Use the boot menu and separate disks...

Lorys89:
> Continuo a non capire perché usi OpenCore per avviare Windows. È solo stupido.
> 
> Usa il menu di avvio e dischi separati ...

and a bootloader that has dual boot, why should I use boot windows from bios? if adding the method _STA resolves? 
it makes no sense what you say.

1alessandro1:
Contional OS patching has been demonstrated to be unreliable, that's why OpenCore does not have policies about when it's necessary to apply the settings described in the config.plist or not.

Just does.

So why are you applying random _STA here and there if booting from the UEFI boot menu is a much cleaner approach to keep OSes and their relative firmwares/drivers separate?

Lorys89:
with your statement, you consider useless the methods is already present in ssdt ec, pmc etc etc
I do not share it, but I accept your thoughts. I await the opinion of the devs.

vit9696:
Makes sense to me, we did not have that on our devices. Could you include the same change for CFL table nearby?

Lorys89:
> Makes sense to me, we did not have that on our devices. Could you include the same change for CFL table nearby?

do you want to apply change to ssdt pnlf and ssdt pnlf cfl? in both?

vit9696:
Correct.

Lorys89:
> Correct.

ok, also applied to ssdt pnlf

Manual/SSDT-PNLF.dsl
```diff
@@ -46,7 +46,17 @@ DefinitionBlock("", "SSDT", 2, "ACDT", "PNLF", 0)
         // 19: CoffeeLake 0xffff
         // 99: Other (requires custom AppleBacklightInjector.kext/WhateverGreen.kext)
         Name(_UID, 0)
-        Name(_STA, 0x0B)
+        Method (_STA, 0, NotSerialized)
+            {
```
stevezhengshiqi, line 50: Thank you for this. Would you mind adjusting the indentation here?

Lorys89 (commented):
fix identation

Lorys89:
> Correct.

do I need to do anything else?

vit9696:
The change looks good to me. Will merge when we get CI fixed.

Lorys89:
> Il cambiamento mi sembra buono. Si fonderanno quando avremo riparato CI.

ok thank you,, but I have ice lake laptop, and the igpu doesn't work well, I can't patch hdmi, and problems with backlight and maximum bkl value, depends on weg, after 3 minutes that start, the backlight turns on, depends on weg.

idalin:
> > Il cambiamento mi sembra buono. Si fonderanno quando avremo riparato CI.
> 
> ok thank you,, but I have ice lake laptop, and the igpu doesn't work well, I can't patch hdmi, and problems with backlight and maximum bkl value, depends on weg, after 3 minutes that start, the backlight turns on, depends on weg.

https://github.com/acidanthera/WhateverGreen/pull/76   try this.

Lorys89:
> > > Il cambiamento mi sembra buono. Si fonderanno quando avremo riparato CI.
> > 
> > 
> > ok thank you,, but I have ice lake laptop, and the igpu doesn't work well, I can't patch hdmi, and problems with backlight and maximum bkl value, depends on weg, after 3 minutes that start, the backlight turns on, depends on weg.
> 
> #76 try this.

thank you, later I fill in and try 😘

Lorys89:
> > > Il cambiamento mi sembra buono. Si fonderanno quando avremo riparato CI.
> > 
> > 
> > ok thank you,, but I have ice lake laptop, and the igpu doesn't work well, I can't patch hdmi, and problems with backlight and maximum bkl value, depends on weg, after 3 minutes that start, the backlight turns on, depends on weg.
> 
> #76 try this.

thank you so much, work well backlight right away, I don't have to wait 3 minutes, great pr your 💪

PR #77, opened 2020-11-29
Add zh_CN translation for README

lulujyc:


Andrey1970AppleLife:
CC @PMheart

vit9696:
I do not believe there is a necessity, and maintaining this will cause extra effort. Thank you!

PR #76, opened 2020-11-27
backlight patch for icl

idalin:
Copied codes from bkl patch for coffee and kaby and applied to icl

PR #78, opened 2020-12-10
IGFX: Convert existing patches to submodules

0xFireWolf:
#### Abstract:

This PR converts more existing IGFX patches to submodules, revises the backlight fix to support ICL platforms, and introduces the concept of coordinated injections.  

#### Coordinated Injections

**Introduction:**
The *Coordinated Injections* module allows multiple submodules to inject code into a shared function. We classify common patterns of code injection as three categories, *prologue injection*, *replacer injection* and *epilogue injection*. The prologue injector injects code before calling the original function, while the epilogue one injects code after calling the original function. The replacer injector overwrites the original implementation and thus does not call the original function. As such, the coordinator maintains three lists of injection descriptors. 

An injection descriptor describes how to inject the code. It defines a trigger value that will be monitored by the coordinator, an injector function that will be called by the coordinator when the trigger value is observed, and a pointer to the next element in the linked list. Both trigger and injector function type are defined by a concrete submodule. Take the shared `ReadRegister32()` submodule as an example, we define the trigger as the register address, so if the graphics driver tries to read a value from that register, the coordinator would invoke the injector function. As a result, we can separate a giant wrapper function into multiple tiny ones where each of them deals with a specific register.

**Implementation:**
Submodules inherited from the `InjectionCoordinator` class get support of coordinated injections automatically, and patch developers do not need write any additional code. Currently, both `ReadRegister32()` and `WriteRegister32()` implement this feature and expose convenient helper functions to access the original implementation. Submodules that require access to these two functions must set their `requiresMMIORegistersReadAccess` and `requiresMMIORegistersWriteAccess` properties to `true` respectively. 

**Related Changes:**
- The backlight registers fix has been revised to use shared MMIO registers-related functions elegantly. It replaces the previous Coffee Lake backlight patch and now supports Intel Ice Lake platforms. As a result, the pull request https://github.com/acidanthera/WhateverGreen/pull/76 is no longer needed.
- A new shared submodule is added to allow other ones to access the global framebuffer controller instance. Similarly, submodules that require access to the global controller must set `requiresGlobalFramebufferControllersAccess` to `true`. As a result, all TODO items proposed in the pull request https://github.com/acidanthera/WhateverGreen/pull/72#discussion_r506913581 are now completed. 

**Future Changes:**
We will probably need to separate AUX registers-related functions into separate shared submodules as well. The maximum link rate fix is the only one that injects code to the `ReadAUX()` function. If separations are necessary, I will probably submit another PR along with a new fix that relies on it.

Thanks for your time,
FireWolf

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 WhateverGreen Changelog
 =======================
+#### v1.4.6
+- Backlight registers fix replaces the previous Coffee Lake backlight fix and is now available on Intel Ice Lake platforms.
+- Boot argument `igfxcflbklt=1` as well as device property `enable-cfl-backlight-fix` are deprecated and replaced by `-igfxblr` and `enable-backlight-registers-fix`.
+
 #### v1.4.5
 - Enabled loading in safe mode (mainly for AGDP fixes)
 - Resolved an issue that the maximum link rate fix is not working properly on Intel Comet Lake platforms. (Thanks @CoronaHack)
--- a/README.md
+++ b/README.md
@@ -69,7 +69,6 @@ Read [FAQs](https://github.com/acidanthera/WhateverGreen/blob/master/Manual/) an
 - `-cdfon` (and `enable-hdmi20` property) to enable HDMI 2.0 patches.
 - `-igfxdump` to dump IGPU framebuffer kext to `/var/log/AppleIntelFramebuffer_X_Y` (available in DEBUG binaries).
 - `-igfxfbdump` to dump native and patched framebuffer table to ioreg at IOService:/IOResources/WhateverGreen
-- `igfxcflbklt=1` boot argument (and `enable-cfl-backlight-fix` property) to enable CFL backlight patch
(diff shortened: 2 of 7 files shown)


Lorys89:
@0xFireWolf

for the problem that the ice lake igpu patch does not start recovery? do you have any feedback on this problem?

m0d16l14n1:
> @0xFireWolf
> 
> for the problem that the ice lake igpu patch does not start recovery? do you have any feedback on this problem?
> photo

In my case (Ice Lake laptop too), as i have already mentioned, i have no difference with that backlight fix / or that new changes. Laptop booting okay, no problems, except that black screen bug for 7-10 seconds on login screen.

vit9696:
Does not seem to have really obvious flaws and contains fairly important changes. Merging as is, let's get it tested for longer time before the release. Thank you.

Lorys89:
> Does not seem to have really obvious flaws and contains fairly important changes. Merging as is, let's get it tested for longer time before the release. Thank you.

Schermata 2020-12-14 alle 21 38 32
Schermata 2020-12-14 alle 21 44 25
to real ioreg mbp 16,2

Lorys89:
@0xFireWolf https://github.com/acidanthera/bugtracker/issues/2241 On beta 13.3 recurs problem of backlighting

PR #79, opened 2020-12-16
IGFX: Add max pixel clock override submodule

al3xtjames:
This can be used as an alternative to patching CoreDisplay.framework. Tested on KBL, but should work on other platforms.


al3xtjames:
Hm, I wonder if we should keep `max-pixel-clock-frequency` configurable - we could just set it to UINT64_MAX and allow any pixel clock. What do you think?

vit9696:
Better not, just in case there is no fool protection. I would say the guides need the values written somewhere. CC @khronokernel

CryptoNeverSleeps:
Quick question... I have XPS 7590 OLED - When resuming from sleep  (Black Screen)... Would this fix that issue.... Only other Alternative is to replace my laptop 4K OLED Screen with a 4K Touch IPS LCD..

al3xtjames:
Probably not, that seems like a different issue.

PR #85, opened 2021-05-19
Update FAQ.Intel.en.md on Haswell graphics

1alessandro1:
Only the HD4400 needs to spoof the `device-id`, so the wiki was updated accordingly

This is mentioned here: https://dortania.github.io/GPU-Buyers-Guide/modern-gpus/intel-gpu.html#haswell-4xxx

and also confirmed by khronokernel here: 

Screenshot 2021-05-19 at 17 47 04

### DIFF
--- a/Manual/FAQ.IntelHD.en.md
+++ b/Manual/FAQ.IntelHD.en.md
@@ -739,7 +739,7 @@ Mobile: 1, PipeCount: 3, PortCount: 1, FBMemoryCount: 1
 - Empty Framebuffer:
   - `0x04120004` (default)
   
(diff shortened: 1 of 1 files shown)


Andrey1970AppleLife:
For mobile HD4600 needs to spoof the device-id

1alessandro1:
@Andrey1970AppleLife Okay, I was confused by the *and*, what about separating the two sentences like this?

For desktop HD4400 *only* and for mobile HD4200/HD4400/HD4600 need fake the device-id 12040000 for IGPU.

Andrey1970AppleLife:
clear without this clarification

PR #87, opened 2021-08-09
Add the backlight smoother for Intel IVB+ platforms

0xFireWolf:
#### Abstract

This pull request contains 1 new feature, 2 enhancements, 2 code relocations, and 1 project setting change.

#### Backlight Smoother

Backlight Smoother (BLS) allows one to adjust the panel brightness smoothly and is available on Intel Ivy Bridge and later. It uses a simple, fast algorithm to change the brightness register gradually toward the value requested by the graphics driver. 

BLS can coexist with the Backlight Registers Fix (BLR) submodule. When both of them are enabled, the MMIO Register Write (RWS) submodule always passes the control to BLR who will then pass the rescaled value to BLS, allowing a chain modification to the same register. When BLR is disabled, RWS passes the control to BLS directly.

#### MMIO Register Access

Both MMIO Register Read and Write submodules are extended to support Intel Ivy Bridge, Haswell, and Broadwell platforms as well, because the controller class names are not `AppleIntelFramebufferController` on these platforms.

#### Relocate Kext Definitions

Kext definitions are relocated to a separate file `kern_igfx_kexts.hpp/cpp`, because both BLR and BLS submodules need to figure out which functions to route based on the framebuffer kext in use instead of the CPU generation.

#### Relocate BLR Implementation

Implementation of BLR and BLS submodules are placed in `kern_igfx_backlight.hpp/cpp`, a dedicated file to put brightness related fixes.

#### Misc Utilities

`kern_igfx_util.hpp/cpp` contains some helper functions and classes needed by the BLS submodule. Let me know if we should put them somewhere else, e.g., Lilu's `kern_util.hpp`.

#### Xcode Project Settings

Switch the C++ standard from `c++1y` to `c++17`, because both `Value::isOneOf()` and `Value::isNotOneOf()` defined in `kern_igfx_util.hpp` rely on C++17's [fold expression](https://en.cppreference.com/w/cpp/language/fold), otherwise Xcode generates a warning.

Thank you for your time.

-- FireWolf
Aug 9, 2021


m0d16l14n1:
Thank you! This one is properly working on my Ice Lake machine now. Previously [AppleBacklightSmoother](https://github.com/hieplpvip/AppleBacklightSmoother) had the issue after waking from sleep > screen is very dim

vit9696:
Thank you very much for this PR. It will take me some time to review, but I believe we can make it into September, 1st release. I do not mind bumping C++ to C++17, but I suggest moving the util stuff to Lilu, as it feels more generally useful. Perhaps it can be put to a separate SDK header, so that we can keep older plugins with C++1y without the need to update their settings.

Emanuele-1998:
> #### Abstract
> This pull request contains 1 new feature, 2 enhancements, 2 code relocations, and 1 project setting change.
> 
> #### Backlight Smoother
> Backlight Smoother (BLS) allows one to adjust the panel brightness smoothly and is available on Intel Ivy Bridge and later. It uses a simple, fast algorithm to change the brightness register gradually toward the value requested by the graphics driver.
> 
> BLS can coexist with the Backlight Registers Fix (BLR) submodule. When both of them are enabled, the MMIO Register Write (RWS) submodule always passes the control to BLR who will then pass the rescaled value to BLS, allowing a chain modification to the same register. When BLR is disabled, RWS passes the control to BLS directly.
> 
> #### MMIO Register Access
> Both MMIO Register Read and Write submodules are extended to support Intel Ivy Bridge, Haswell, and Broadwell platforms as well, because the controller class names are not `AppleIntelFramebufferController` on these platforms.
> 
> #### Relocate Kext Definitions
> Kext definitions are relocated to a separate file `kern_igfx_kexts.hpp/cpp`, because both BLR and BLS submodules need to figure out which functions to route based on the framebuffer kext in use instead of the CPU generation.
> 
> #### Relocate BLR Implementation
> Implementation of BLR and BLS submodules are placed in `kern_igfx_backlight.hpp/cpp`, a dedicated file to put brightness related fixes.
> 
> #### Misc Utilities
> `kern_igfx_util.hpp/cpp` contains some helper functions and classes needed by the BLS submodule. Let me know if we should put them somewhere else, e.g., Lilu's `kern_util.hpp`.
> 
> #### Xcode Project Settings
> Switch the C++ standard from `c++1y` to `c++17`, because both `Value::isOneOf()` and `Value::isNotOneOf()` defined in `kern_igfx_util.hpp` rely on C++17's [fold expression](https://en.cppreference.com/w/cpp/language/fold), otherwise Xcode generates a warning.
> 
> Thank you for your time.
> 
> -- FireWolf
> Aug 9, 2021

This new integration is very useful, thanks. I have an idea, can you enter a minimum brightness level to keep it from reaching 0 and therefore with a completely black screen?

0xFireWolf:
> Thank you very much for this PR. It will take me some time to review, but I believe we can make it into September, 1st release. I do not mind bumping C++ to C++17, but I suggest moving the util stuff to Lilu, as it feels more generally useful. Perhaps it can be put to a separate SDK header, so that we can keep older plugins with C++1y without the need to update their settings.

Thank you. Please take your time.

I have moved those helpers to Lilu's `kern_util.hpp/cpp`. `Value::isOneOf()` and `Value::isNotOneOf()` are now guarded by a language version check, so we don't have to create a new file in Lilu.

0xFireWolf:
> This new integration is very useful, thanks. I have an idea, can you enter a minimum brightness level to keep it from reaching 0 and therefore with a completely black screen?

@Emanuele-1998 So I guess the lowest brightness level on your laptop is a black screen (which actually is a feature of your display panel 😀). Yes, it is possible to specify the lower and/or upper bound of the register value via device properties, but do we really need it?

Emanuele-1998:
> > This new integration is very useful, thanks. I have an idea, can you enter a minimum brightness level to keep it from reaching 0 and therefore with a completely black screen?
> 
> 
> 
> @Emanuele-1998 So I guess the lowest brightness level on your laptop is a black screen (which actually is a feature of your display panel 😀). Yes, it is possible to specify the lower and/or upper bound of the register value via device properties, but do we really need it?

why do you say it is a feature of my display panel? I have not yet seen laptops that do not expect if you pull the brightness to 0 at least only for the lower limit. It might be useful to specify this to have a kext as complete as possible. also how could this thing be implemented / put in the device properties, ie which attributes in the properties go inside?

Emanuele-1998:
> > This new integration is very useful, thanks. I have an idea, can you enter a minimum brightness level to keep it from reaching 0 and therefore with a completely black screen?
> 
> 
> 
> @Emanuele-1998 So I guess the lowest brightness level on your laptop is a black screen (which actually is a feature of your display panel 😀). Yes, it is possible to specify the lower and/or upper bound of the register value via device properties, but do we really need it?

why do you say it is a feature of my display panel? I have not yet seen laptops that do not have a black screen if you set the brightness to 0. It may be useful to specify a lower limit in order to have as complete a kext as possible. also how could this thing be implemented / inserted in the device properties, ie what attributes in the properties go inside?

0xFireWolf:
> Why do you say it is a feature of my display panel? I have not yet seen laptops that do not have a black screen if you set the brightness to 0. 

My screen does not go black when the driver writes 0 to the PWM frequency register (which corresponds to the lowest brightness level), regardless of which operating system is running. The panel remains dim but I can still see my desktop.

I guess we are not talking about `0` under the same context, so I could be wrong. I understand that some people like you just want to prevent the screen from going black, so I will add an option for you guys to specify the range.

Emanuele-1998:
> > Why do you say it is a feature of my display panel? I have not yet seen laptops that do not have a black screen if you set the brightness to 0.
> 
> My screen does not go black when the driver writes 0 to the PWM frequency register (which corresponds to the lowest brightness level), regardless of which operating system is running. The panel remains dim but I can still see my desktop.
> 
> I guess we are not talking about `0` under the same context, so I could be wrong. I understand that some people like you just want to prevent the screen from going black, so I will add an option for you guys to specify the range.

my screen compared to others remains completely black without showing my desktop. While others remain black by showing the desktop, so it will be useful to establish a range for both cases. Thanks for the work

0xFireWolf:
> My screen compared to others remains completely black without showing my desktop. While others remain black by showing the desktop, so it will be useful to establish a range for both cases. Thanks for the work you are doing 

@Emanuele-1998 I have updated the code. Please read the manual and give a try. I used this new feature to slightly increase the brightness at the lowest level, and it worked well on my machine. Feel free to report any issues.

dreamwhite:
Good morning @0xFireWolf, I tried manually compiling your latest commits but the compiler keeps yelling at me about `BrightnessRequestQueue` and `CircularBuffer` template missing. Are you aware of this stuff or eventually can you give us a working artifact?
Many thanks and have a great day

Ab2774:
Hi @0xFireWolf, first of all, thank you so much for your work, a lot of people including me were waiting for this moment to come, AppleBacklightSmoother hasn't been updated for almost a year, but this pull request has a lot of customizable features unlike AppleBacklightSmoother, and it finally solved the black screen issue, anyway, I just encountered an issue, which is setting both `backlight-smoother-steps` and `backlight-smoother-interval` will cause a Kernel Panic, I don't know if that's expected, if so, then I suggest adding a fail-safe.
aside from that, it's all working well, still haven't found the best value for `backlight-smoother-steps` and `backlight-smoother-interval` for my Kaby Lake laptop, by default, the transition is smooth but slow, I will try in the next few days.
Thanks

0xFireWolf:
@dreamwhite Please check the [manual](https://github.com/0xFireWolf/WhateverGreen/blob/master/Manual/FAQ.IntelHD.en.md#customize-the-behavior-of-the-backlight-smoother-to-improve-your-experience).

@Ab2774 While it is fine to set the interval to 0 milliseconds, it does not make any sense to set the number of steps to 0. Transitioning to the target brightness level requires at least a single step, which is equivalent to skipping the smoother. I will sanitize the value in case users provide an invalid one unintentionally. If you find the transition slow when you press the brightness keys, try to reduce the `interval` value to 5 ms and `steps` to 20 or 25.

Ab2774:
Hi @0xFireWolf, thanks for the feedback, I tried to set the `interval` and the `steps` to `20` or `25` as you mentioned, and it's not slow anymore, except, there's no smooth transition, so, after a few experiments, I found that setting the `interval` to `9` and the `steps` to `20` seems good, it's still slow but not as slow as before, and the transition is smooth.
Also, I just noticed, unlike AppleBacklightSmoother, when setting my laptop to sleep from the Menu Bar, there's no transition, and waking the laptop from sleep (black screen) to the lock screen or to the desktop, there's also no transition.
Thanks

WhateverGreen/kern_igfx.hpp
```diff
@@ -1590,6 +1585,143 @@ class IGFX {
...
+
+		/**
+		 *  [CFL+] Wrapper to write to BXT_BLC_PWM_DUTY1 smoothly
+		 *
+		 *  @note When this function is called, `address` is guaranteed to be `BXT_BLC_PWM_DUTY1`.
+		 */
+		static void smoothCFLWriteRegisterPWMDuty1(void *controller, uint32_t address, uint32_t value);
+
+		/**
+		 *  [IVB*] A replacer descriptor that injects code when the register of interest is BLC_PWM_CPU_CTL
```
vit9696, line 1703: Same here? IVB+?

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -0,0 +1,452 @@
...
+		callbackIGFX->modMMIORegistersWriteSupport.replacerList.add(&dCFLPWMFreq1);
+		callbackIGFX->modMMIORegistersWriteSupport.replacerList.add(&dCFLPWMDuty1);
+	} else {
+		SYSLOG("igfx", "BLS: [ERR!] Found an unsupported platform. Will not perform any injections.");
+	}
+}
+
+void IGFX::BacklightRegistersFix::wrapKBLWriteRegisterPWMFreq1(void *controller, uint32_t reg, uint32_t value) {
+	DBGLOG("igfx", "BLR: [KBL*] WriteRegister32<BXT_BLC_PWM_FREQ1>: Called with register 0x%x and value 0x%x.", reg, value);
+	assertf(reg == BXT_BLC_PWM_FREQ1, "Fatal Error: Register should be BXT_BLC_PWM_FREQ1.");
```
vit9696, line 92: Use PANIC_COND.

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -0,0 +1,452 @@
...
+	if (owner == nullptr) {
+		SYSLOG("igfx", "BLS: Failed to create the owner of the event source.");
+		goto error0;
+	}
+	
+	// Initialize the request queue
+	queue = BrightnessRequestQueue::withCapacity(queueSize);
+	if (queue == nullptr) {
+		SYSLOG("igfx", "BLS: Failed to initialize the request queue.");
+		goto error1;
```
vit9696, line 369: I strongly dislike these gotos, they make me feel there needs to be an outer function that will do failure printing and likely cleanup.

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -0,0 +1,452 @@
...
+	
+	// Submit the request and notify the event source
+	callbackIGFX->modBacklightSmoother.queue->push(BrightnessRequest(controller, address, value));
+	callbackIGFX->modBacklightSmoother.eventSource->enable();
+	DBGLOG("igfx", "BLS: [IVB*] WriteRegister32<BLC_PWM_CPU_CTL>: The brightness request has been submitted.");
+}
+
+void IGFX::BacklightSmoother::smoothHSWWriteRegisterPWMFreq1(void *controller, uint32_t address, uint32_t value) {
+	DBGLOG("igfx", "BLS: [HSW+] WriteRegister32<BXT_BLC_PWM_FREQ1>: Called with register 0x%x and value 0x%x.", address, value);
+	assertf(address == BXT_BLC_PWM_FREQ1, "Fatal Error: Register should be BXT_BLC_PWM_FREQ1.");
```
vit9696, line 436: Same here and below, use Lilu PANIC macros.

WhateverGreen/kern_igfx_backlight.hpp
```diff
@@ -0,0 +1,122 @@
...
+static constexpr uint32_t BXT_BLC_PWM_DUTY1 = 0xC8258;
+
+/**
+ *  Represents a single brightness adjustment request
+ */
+struct BrightnessRequest {
+	/**
+	 *  The framebuffer controller
+	 */
+	void *controller;
```
vit9696, line 30: Please add inline zero init for safety reasons everywhere.

vit9696 (commented):
Looks pretty good, marked minor stuff inline.

WhateverGreen/kern_igfx_backlight.cpp
0xFireWolf: I also hate them, but I do find them simplify error handling and clean-up code. Anyway, I revised the implementation of `deinit()` and replaced those `goto` statements with `deinit()` invocations.

WhateverGreen/kern_igfx.hpp
```diff
@@ -1506,32 +1507,26 @@ class IGFX {
...
-		 *  Can be specified via max-backlight-freq property.
+		 *  [COMM] User-requested backlight frequency obtained from BXT_BLC_PWM_FREQ1 at system start.
+		 *
+		 *  @note Its value can be specified via max-backlight-freq property.
 		 */
 		uint32_t targetBacklightFrequency {};
 		
 		/**
-		 *  [KBL] User-requested pwm control value obtained from BXT_BLC_PWM_CTL1.
+		 *  [KBL*] User-requested pwm control value obtained from BXT_BLC_PWM_CTL1.
```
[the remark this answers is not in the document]
0xFireWolf: No, I use the following notations to filter kernel logs more easily when debugging the kext. 

- `[KBL+]` means KBL and later.
- `[KBL*]` means KBL only.
- `[KBL-]` means KBL and earlier.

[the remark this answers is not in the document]
vit9696: Maybe [KBL] then? * really makes me confused, I am also unsure it is any common.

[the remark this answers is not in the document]
0xFireWolf: Done.

PR #95, opened 2021-10-21
Always execute the latest brightness change request instead of queuing

usr-sse2:


usr-sse2:
I wanted not to remove `sin` but to calculate all necessary values once and save them into an array.
I think we don't need to support devices with more than one built-in LVDS display.

vit9696:
Right, perhaps we can put the precalculation to a separate PR and merge this code as is for now? Could you update the title and include the changelog changes we need to apply after merging?

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -357,6 +374,17 @@ void IGFX::BacklightSmoother::processKernel(KernelPatcher &patcher, DeviceInfo *
 		brightnessRange.second = UINT32_MAX;
 	}
 	
+	sines = static_cast<double*>(kern_os_malloc(steps * sizeof(double)));
```
vit9696, line 377: Use Buffer::create<double>, in my opinion. It is specifically for that purpose.

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -239,26 +240,39 @@ bool BrightnessRequestEventSource::checkForWork() {
...
 				IGFX::callbackIGFX->writeRegister32(request.controller, request.address, request.getTargetRegisterValue(value));
 				IOSleep(smoother->interval);
+				if (smoother->request.id != request.id) {
+					return true; // Start again, because there is a new request
+				}
 			}
 		} else if (cbrightness > tbrightness) {
 			// Decrease the brightness
-			for (uint32_t value = cbrightness - stride; value >= tbrightness + stride; value -= stride) {
+			for (uint32_t i = 1; i < smoother->steps; i++) {
```
vit9696, line 269: Leave a comment that 1 is intended here.

usr-sse2:
Don't merge this now.
1. The formula is incorrect, should be (1-cos(π*i/steps))/2 so that the derivative is 0 at both ends of the interval.
2. Even with this replacement, this function is worse than linear when a "Brightness Up" or "Brightness Down" key is pressed  and held, so it seems that it's better to revert to linear function.

WhateverGreen/kern_igfx_backlight.cpp
usr-sse2: Removed

WhateverGreen/kern_igfx_backlight.cpp
usr-sse2: Done

Craze1666 (approved):

PR #97, opened 2022-01-13
[Suggestions] [Experimental] Ice Lake backlight fix.

preflight-f:
Related Issues:

- https://github.com/acidanthera/bugtracker/issues/1256

Tested on Surface Laptop Go.

Short explanations:

- CFL: wrapCFLWriteRegisterPWMFreq1 -> wrapCFLWriteRegisterPWMDuty1
- ICL: wrapCFLWriteRegisterPWMDuty1 -> LightUpEDP (might not be called at startup) -> wrapCFLWriteRegisterPWMFreq1 -> ...

### DIFF
--- a/WhateverGreen/kern_igfx_backlight.cpp
+++ b/WhateverGreen/kern_igfx_backlight.cpp
@@ -184,6 +184,18 @@ void IGFX::BacklightRegistersFix::wrapCFLWriteRegisterPWMDuty1(void *controller,
 	DBGLOG("igfx", "BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Called with register 0x%x and value 0x%x.", reg, value);
 	PANIC_COND(reg != BXT_BLC_PWM_DUTY1, "igfx", "Fatal Error: Register should be BXT_BLC_PWM_DUTY1.");
 	
+	// [Experimental] Ice Lake backlight fix.
+	// TODO: Excluding CFL (Is it really necessary?)
+
+	if (value && callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency == 0) {
+		uint32_t registerValue = callbackIGFX->readRegister32(controller, 0xC2014);
+
+		uint32_t freq = 17777;
+  		if (registerValue & (1 << 8)) freq = 22222;
+
+		wrapCFLWriteRegisterPWMFreq1(controller, BXT_BLC_PWM_FREQ1, freq);
+	}
+
 	if (callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency && callbackIGFX->modBacklightRegistersFix.targetBacklightFrequency) {
 		// Translate the PWM duty cycle between the driver scale value and the HW scale value
 		uint32_t rescaledValue = static_cast<uint32_t>((value * static_cast<uint64_t>(callbackIGFX->modBacklightRegistersFix.targetBacklightFrequency)) / static_cast<uint64_t>(callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency));
--- a/WhateverGreen/kern_igfx_backlight.cpp
+++ b/WhateverGreen/kern_igfx_backlight.cpp
@@ -185,15 +185,13 @@ void IGFX::BacklightRegistersFix::wrapCFLWriteRegisterPWMDuty1(void *controller,
 	PANIC_COND(reg != BXT_BLC_PWM_DUTY1, "igfx", "Fatal Error: Register should be BXT_BLC_PWM_DUTY1.");
 	
 	// [Experimental] Ice Lake backlight fix.
-	// TODO: Excluding CFL (Is it really necessary?)
 
 	if (value && callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency == 0) {
-		uint32_t registerValue = callbackIGFX->readRegister32(controller, 0xC2014);
-
-		uint32_t freq = 17777;
-  		if (registerValue & (1 << 8)) freq = 22222;
-
-		wrapCFLWriteRegisterPWMFreq1(controller, BXT_BLC_PWM_FREQ1, freq);
+		SYSLOG("igfx", "BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Ice Lake backlight fix was entered.");
+		uint32_t registerValue = callbackIGFX->readRegister32(controller, SFUSE_STRAP);
+		uint32_t selectedFreq = ICL_FREQ_NORMAL;
+  		if (registerValue & SFUSE_STRAP_RAW_FREQUENCY) selectedFreq = ICL_FREQ_RAW;
+		wrapCFLWriteRegisterPWMFreq1(controller, BXT_BLC_PWM_FREQ1, selectedFreq);
 	}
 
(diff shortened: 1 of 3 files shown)


vit9696:
CC @0xFireWolf @usr-sse2

0xFireWolf:
It took me a while to understand what you are doing here. 
It would be greatly appreciated if you elaborate on the changes you have made in your pull request. : )

So let's begin with a quick recap on the backlight issue on Ice Lake platforms.
Previously, the revised Backlight Register Fix (BLR) submodule (https://github.com/acidanthera/WhateverGreen/pull/78) added support for ICL platforms to avoid the 3-minute black screen but did not solve the issue that the maximum brightness under macOS was lower than that under other operating systems on some laptops. As far as I know, some Lenovo laptops have this issue. 

I assume that your code snippet that sets Apple's hard-coded frequency value is extracted from the `AppleIntelFramebufferController::start()` function. The code snippet is not new as it is in the CFL driver as well. In the `start()` function, Apple's driver chooses which frequency to use based upon whether the bit `SFUSE_STRAP_RAW_FREQUENCY = (1 << 8)` is set in the register `SFUSE_STRAP` at `0xc2014`, and it then stores the frequency to an instance variable of the controller. Later, the driver fetches the stored frequency and writes it to the register `BXT_BLC_PWM_FREQ1` at `0xC8254` in the `LightUpEDP()` function. However, depending upon the hardware CRTC state set by the firmware, the driver may complete the modeset without invoking `LightUpEDP()` at boot time. As a result, when a user requests to change the brightness, WEG may not scale the value of the duty register.

So that's my interpretation of the change you have made. You basically sets the frequency manually if the driver does not invoke the light up function before changing the brightness level. Please feel free to correct me if my interpretation is wrong. 

Now I would like to confirm the following things.
- Did you see the line `BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Write PWM_DUTY1 has zero frequency driver` in the kernel log without your changes?
- Does this change fix the low brightness issue on **ALL** ICL laptops?
- I am not sure whether we need this change for CFL platforms as well.

Perhaps, we need to invite people who own ICL laptops to test the patch before merging it to the main branch.
Incidentally, please add the register and the bit field definition in `kern_igfx_backlight.hpp` instead of using them as magic values in the code.

0xFireWolf:
> Yes. But I'm not sure filling the same value as Apple did is reasonable.

It looks fine to me, because WEG needs the frequency requested by Apple's driver to rescale the value of the duty register. 

> > Did you see the line BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Write PWM_DUTY1 has zero frequency driver in the kernel log without your changes?
> 
> Yes.

As expected, otherwise it does not make sense.

> > Does this change fix the low brightness issue on ALL ICL laptops?
> 
> I cannot confirm it, for the shortage of ICL laptops.

No worries. I mean it would be better to invite other people to test your patch as well.

> > I am not sure whether we need this change for CFL platforms as well.
> 
> The driver of CFL will never match `value && callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency == 0` so I thought we can keep that if-block for both.
> But I'm still not sure.

Well, I think it does not hurt if we keep that for CFL as well, because the CFL driver uses the same hard-coded frequencies and it is possible that `LightUpEDP()` is not invoked on some CFL laptops at boot.

@vit9696 This fix makes sense to me, and I suggest to merge it, but it would be better if we can get some feedback from other folks. What's your suggestion?

WhateverGreen/kern_igfx_backlight.cpp
```diff
@@ -184,6 +184,16 @@ void IGFX::BacklightRegistersFix::wrapCFLWriteRegisterPWMDuty1(void *controller,
 	DBGLOG("igfx", "BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Called with register 0x%x and value 0x%x.", reg, value);
 	PANIC_COND(reg != BXT_BLC_PWM_DUTY1, "igfx", "Fatal Error: Register should be BXT_BLC_PWM_DUTY1.");
 	
+	// [Experimental] Ice Lake backlight fix.
+
+	if (value && callbackIGFX->modBacklightRegistersFix.driverBacklightFrequency == 0) {
+		SYSLOG("igfx", "BLR: [CFL+] WriteRegister32<BXT_BLC_PWM_DUTY1>: Ice Lake backlight fix was entered.");
```
vit9696, line 190: Does that really need to be SYSLOG instead of DBGLOG?

vit9696:
I believe we should merge this too and wait for the reports. If anyone gets issues with it, I will CC. Perhaps we can make it configurable at worst.

0xFireWolf:
> I believe we should merge this too and wait for the reports. If anyone gets issues with it, I will CC. Perhaps we can make it configurable at worst.

Sounds like a good plan. 
We can add a boot argument to enable this fix and leave it there for a month or two. 
If everything goes well, we then remove the boot argument from the future releases.

preflight-f:
@0xFireWolf That’s nice. I’ll do that later. (Adding -igfxblrexpr)

vit9696:
Nah, better enable by default. And then disable on reports. Otherwise nobody will enable it.

preflight-f:
@vit9696 @0xFireWolf Changes were made. Now it can be merged.

vit9696:
Errng, like I said, no need for an argument right now. Better introduce it in a subsequent commit if there are issues. It will require a device property check as well anyway.

Lorys89:
ice lake users can run tests on chat gitter https://gitter.im/ICE-LAKE-HACKINTOSH-DEVELOPMENT

preflight-f:
I thought similarly to @0xFireWolf that we need more usage reports before making it as a default behavior.

vit9696:
You will not get usage reports from the people that do not need this flag, because they have no reasons to set it. And we need the opinions of these users most to determine whether this flag can be enabled by default. There is no better way to get this opinion other than force them having this feature and then wait for issues.

preflight-f:
The argument was removed now.

Lorys89:
> The argument was removed now.

in my ice lake DELL laptop, there is no change or improvement,it worked before, and it works now.

hopefully on other laptops it will solve the problems.

CobanRamo:
I can confirm that on my Lenovo S340 IceLake (81VW00C4GE) it finally solved the problem with dimmed backlight. The backlight was only after the sleep fully on Windows level.
The latest Whatevergreen works fine for me with Catalina, BigSur and Monterey.

Lorys89:
@0xFireWolf  I wanted to ask you a question, I use ig platform 0200518a and everything works fine, if I use 0200538a comes to black screen desk, but ariva safe to desk pouche if I turn up and down the volume I hear the sounds of macos, ig platform 0200538a is the one used natively by mbp 16.2
your opinion about it?

TheHackGuy:
> @0xFireWolf  I wanted to ask you a question, I use ig platform 0200518a and everything works fine, if I use 0200538a comes to black screen desk, but ariva safe to desk pouche if I turn up and down the volume I hear the sounds of macos, ig platform 0200538a is the one used natively by mbp 16.2
> your opinion about it? 

Different platform id's have different connectors, busids, flags, etc. It's pretty much normal one will work and another one doesn't.

PR #102, opened 2022-07-07
Remove profile 2 from VTSupportedProfileArray

PMheart:
In addition to the removal of vp9 flag `IOGVAXDecode` on SKL when spoofing as KBL, profile 2 under `VTSupportedProfileArray` in `IOGVAHEVCDecodeCapabilities` also needs to be removed to avoid flickering when decoding HEVC encoded videos.

CC and Thanks to @dhinakg for the code infrastructure again!

### DIFF
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -1000,7 +1000,7 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
 	if (callbackIGFX->disableAccel)
 		return false;
 	
-	OSDictionary* developmentDictCpy {};
+	OSDictionary *developmentDictCpy {};
 
 	if (callbackIGFX->fwLoadMode != FW_APPLE || callbackIGFX->modForceWakeWorkaround.enabled) {
 		auto developmentDict = OSDynamicCast(OSDictionary, that->getProperty("Development"));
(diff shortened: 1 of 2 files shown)


WhateverGreen/kern_igfx.cpp
```diff
@@ -1072,8 +1072,32 @@ bool IGFX::wrapAcceleratorStart(IOService *that, IOService *provider) {
 		that->setName("IntelAccelerator");
 	
 	if (callbackIGFX->forceSKLAsKBL) {
-		DBGLOG("igfx", "disabling VP9 hw decode support on Skylake when using KBL kexts");
+		DBGLOG("igfx", "disabling VP9 hw decode support on Skylake with KBL kexts");
 		that->removeProperty("IOGVAXDecode");
+
+		if (auto hevcDecodeCap = OSDynamicCast(OSDictionary, that->getProperty("IOGVAHEVCDecodeCapabilities"))) {
```
vit9696, line 1078: That's C++17, this feature is cool, but it does not make the code more readable when you make so many nested levels. Could you please move it to a new function and instead return from it on failing to get the property or allocate a dictionary?

PMheart: https://github.com/acidanthera/WhateverGreen/commit/a9edfe199550a37c0f1172f836a9d18dd09919eb

WhateverGreen/kern_igfx.cpp
```diff
@@ -996,11 +996,55 @@ OSObject *IGFX::wrapCopyExistingServices(OSDictionary *matching, IOOptionBits in
 	return FunctionCast(wrapCopyExistingServices, callbackIGFX->orgCopyExistingServices)(matching, inState, options);
 }
 
+bool IGFX::removeVTSupportedProfile2(IOService *that) {
+	if (!that)
```
vit9696, line 1000: Makes no sense to check this, as it is dereferenced right before calling this function.

WhateverGreen/kern_igfx.cpp
```diff
@@ -996,11 +996,55 @@ OSObject *IGFX::wrapCopyExistingServices(OSDictionary *matching, IOOptionBits in
...
+
+	auto vtSuppProfCpy = OSArray::withArray(vtSuppProf);
+	if (!vtSuppProfCpy) {
+		hevcDecodeCapCpy->release();
+		return false;
+	}
+
+	auto count = vtSuppProfCpy->getCount();
+	bool found = false;
+	for (auto i = 0; i < count; i++) {
```
vit9696, line 1025: auto here will default to int, better use size_t for both count and i.

WhateverGreen/kern_igfx.cpp
```diff
@@ -996,11 +996,55 @@ OSObject *IGFX::wrapCopyExistingServices(OSDictionary *matching, IOOptionBits in
...
+		auto num = OSDynamicCast(OSNumber, vtSuppProfCpy->getObject(i));
+		if (num->unsigned8BitValue() == 2) {
+			DBGLOG("igfx", "removing profile 2 from VTSupportedProfileArray/IOGVAHEVCDecodeCapabilities index %u on Skylake with KBL kexts", i);
```
vit9696, line 1028: Missing null check for num.

WhateverGreen/kern_igfx.cpp
PMheart: https://github.com/acidanthera/WhateverGreen/commit/f9cff542db12c3257068490039a2b8535bc016be

WhateverGreen/kern_igfx.cpp
PMheart: https://github.com/acidanthera/WhateverGreen/commit/1d26c0f33ff6627e07d563bcc96854bdf8b33dd7

WhateverGreen/kern_igfx.cpp
PMheart: `getCount()` returns `unsigned int`, so I rather followed it: https://github.com/acidanthera/WhateverGreen/commit/e6f90ecbcc605c608a2d945793ac7ab706579179

dhinakg (commented):
LGTM, although I wonder whether it is a good idea to move the `IOGVAXDecode` code into the function too and rename it

PMheart:
> LGTM, although I wonder whether it is a good idea to move the `IOGVAXDecode` code into the function too and rename it

https://github.com/acidanthera/WhateverGreen/commit/035cd10ff3492d176d6398e0f2564b4ec242b451

vit9696:
Looks good. Please merge.

PR #103, opened 2022-07-10
Remove profile 2 from IOGVAHEVCEncodeCapabilities

PMheart:
SKL does not support 10-bit hardware encoding/decoding, and this causes freezing when attempting to do so.
The removal of profile 2 under VTSupportedProfileArray allows fallback to software encoding/decoding.
Thanks dhinakg and aben for finding this.

### DIFF
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -1062,39 +1062,40 @@ bool IGFX::applySklAsKblPatches(IOService *that) {
 	// The removal of profile 2 under VTSupportedProfileArray allows fallback to software encoding/decoding.
 	// Thanks dhinakg and aben for finding this.
 	const char *hevcCapProps[] = { "IOGVAHEVCDecodeCapabilities", "IOGVAHEVCEncodeCapabilities" };
+	bool found = false;
 	for (auto prop : hevcCapProps) {
+		found = false;
+
 		auto hevcCap = OSDynamicCast(OSDictionary, that->getProperty(prop));
 		if (!hevcCap)
-			return false;
+			continue;
 
 		auto newHevcCap = OSDictionary::withDictionary(hevcCap);
 		if (!newHevcCap)
-			return false;
+			continue;
 
 		auto vtSuppProf = OSDynamicCast(OSArray, newHevcCap->getObject("VTSupportedProfileArray"));
 		if (!vtSuppProf) {
 			newHevcCap->release();
-			return false;
+			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


WhateverGreen/kern_igfx.cpp
```diff
@@ -1058,47 +1058,55 @@ bool IGFX::applySklAsKblPatches(IOService *that) {
...
-	if (!hevcDecodeCap)
-		return false;
+	// SKL does not support 10-bit hardware encoding/decoding, and this causes freezing when attempting to do so.
+	// The removal of profile 2 under VTSupportedProfileArray allows fallback to software encoding/decoding.
+	// Thanks dhinakg and aben for finding this.
+	const char *hevcCapProps[] = { "IOGVAHEVCDecodeCapabilities", "IOGVAHEVCEncodeCapabilities" };
+	for (auto prop : hevcCapProps) {
+		auto hevcCap = OSDynamicCast(OSDictionary, that->getProperty(prop));
+		if (!hevcCap)
+			return false;
```
vit9696, line 1068: Why false? Maybe continue here and below?

WhateverGreen/kern_igfx.cpp
```diff
@@ -1058,47 +1058,55 @@ bool IGFX::applySklAsKblPatches(IOService *that) {
...
+		newVtSuppProf->release();
 
-	that->setProperty("IOGVAHEVCDecodeCapabilities", newHevcDecodeCap);
-	newHevcDecodeCap->release();
+		that->setProperty(prop, newHevcCap);
+		newHevcCap->release();
+	}
 
-	return found;
+	return true;
```
vit9696, line 1109: Needs a var set to true in the end of the iteration in my opinion.

PMheart: Above. So if everything flows, nothing wrong happens, hence true.

WhateverGreen/kern_igfx.cpp
vit9696: Well, not having decoder or encoder feels ok to me. Apple may update the profiles at their preference.

PMheart: https://github.com/acidanthera/WhateverGreen/commit/02ab31c6263a80571fa18b8c914df03585ce1846

PMheart:
> in my opinion, drop this and return true on at least one removal

https://github.com/acidanthera/WhateverGreen/pull/103/commits/f161a2029e1181d4994b9b700b0f54fa22c932c9

vit9696:
Looks good now.

PR #104, opened 2022-07-19
Set `VTRating` and `IOGVAHEVC` properties to SKL values, idea from @dhinakg

PMheart:


PMheart:
This is not proven to help improve the compatibility with SKLAsKBL spoofing. Closing.

PR #110, opened 2022-12-23
iMac 106b GPU identifiers 2009-2017

nikey22:
MacOS Ventura has dropped official support for imacs up to and including 2015 models. I've paid particular attention to these models as they will now require OCLP/WhateverGreen to be patched.  

I present here additional iMac (106b) GPU ID identifiers pulled from iMac 2009 to iMac 2017 machines. There are also some entries for dGPUs that we have been able to integrate into the iMac machines with a vbios hack.


vit9696:
Thanks!

PR #116, opened 2023-06-25
Extend the Backlight Registers Alternative Fix (BLT) submodule to support Kaby Lake platforms

0xFireWolf:
**Abstract:**

This PR is a follow-up to https://github.com/acidanthera/WhateverGreen/pull/113. The Backlight Registers Alternative Fix (BLT) has been redesigned and now supports both Kaby Lake and Coffee Lake platforms. The original Backlight Registers Fix (BLR) is now compatible with the Backlight Smoother (BLS) when running on Kaby Lake platforms. Previously, it does not pass the re-scaled duty cycle value to the smoother.

**Affected System Versions:**

- macOS Ventura 13.4 (22F66)
- macOS Ventura 13.4.1 (22F82)
- macOS Sonoma 14.0 DP1 (23A5257q)
- macOS Sonoma 14.0 DP2 (23A5276g)

Note that macOS Ventura 13.5 DP1/2/3 are not affected.

**Boot Argument and Device Property:**

- The boot argument `-igfxblt` and the device property `backlight-registers-alternative-fix` remain the same.

Please take your time to review it. Thank you.

-- FireWolf


xiechunjiang:
Intel HD Graphics 630  macOS Sonoma(23A344)  -igfxblt -igfxbls   It doesn't work. When the login interface appears, press the power button for about 2 seconds, then close and open the laptop lid. Then the display will be normal.

PR #120, opened 2024-11-09
Added Alder Lake/Raptor Lake CPU detection

al3xtjames:
I noticed this was missing after 403562b ("Added Arrow Lake CPU detection").


vit9696:
Thanks!

PR #121, opened 2025-02-01
Update kern_igfx.cpp исправленно

sergick2020:
tigerlake+fix=Graphics iris@Xe.

### DIFF
--- a/WhateverGreen/kern_igfx.cpp
+++ b/WhateverGreen/kern_igfx.cpp
@@ -122,6 +122,13 @@ void IGFX::init() {
 			currentFramebufferOpt = &kextIntelICLHPFb;
 			modDVMTCalcFix.available = true;
 			break;
+		case CPUInfo::CpuGeneration::TigerLake:
+			supportsGuCFirmware = true;
+			currentGraphics = &kextIntelICL;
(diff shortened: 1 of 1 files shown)


WhateverGreen/kern_igfx.cpp
```diff
@@ -122,6 +122,13 @@ void IGFX::init() {
 			currentFramebufferOpt = &kextIntelICLHPFb;
 			modDVMTCalcFix.available = true;
 			break;
+		case CPUInfo::CpuGeneration::TigerLake:
```
vit9696, line 125: To be honest, I thought that Ice Lake drivers do not work as is on Tiger Lake hardware. Although I can merge as is, since it will not do harm on systems unsupported as is, could you please confirm it solves any actual issues?

vit9696:
There is work ongoing to enable TGL+ graphics in https://github.com/lshbluesky/WhateverGreen by @lshbluesky and colleagues. This PR definitely does not provide all the necessary changes from there, so I am going to close this. I guess we can try to merge @lshbluesky changes, in case he reads this, but currently there is still no proper support for TGL+ acceleration.

PR #124, opened 2025-09-16
wait for IOGraphicsFamily to load before patching AMDSupport on Tahoe

royalgraphx:
Since Tahoe Developer Beta 1, WhateverGreen causes an Invalid Opcode (6) kernel panic when booting into recoveryOS, and during the update phases. This has been resolved by simply ensuring that IOGraphicsFamily has loaded prior to patching AmdSupport.kext, and I've wrapped the appropriate call to only run on Tahoe, to ensure that users of Sonoma and below, are not affected, without having to test.

Although, I personally have tested on Sonoma 15.6 and Tahoe DB1, and have gotten a confirmed boot on both recoveryOS and during the update phases.

### DIFF
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -41,6 +41,8 @@ static const char *idRadeonX6000New {"com.apple.kext.AMDRadeonX6000"};
 static const char *idRadeonX3000Old {"com.apple.AMDRadeonX3000"};
 static const char *idRadeonX4000Old {"com.apple.AMDRadeonX4000"};
 
+static KernelPatcher::KextInfo kextIOGraphicsFamily
+{ "com.apple.iokit.IOGraphicsFamily", nullptr, 0, {}, {}, KernelPatcher::KextInfo::Unloaded };
 static KernelPatcher::KextInfo kextRadeonFramebuffer
 { "com.apple.kext.AMDFramebuffer", pathFramebuffer, arrsize(pathFramebuffer), {}, {}, KernelPatcher::KextInfo::Unloaded };
 static KernelPatcher::KextInfo kextRadeonLegacyFramebuffer
@@ -79,6 +81,16 @@ static const char *powerGatingFlags[] {
 	"CAIL_DisableAcpPowerGating",
 	"CAIL_DisableSAMUPowerGating"
 };
+
+/**
+ *  This function will be called by Lilu ONLY when IOGraphicsFamily has loaded.
+ *  Its only job is to then activate our patches for AMDSupport.
+ */
+static void onIOGraphicsLoad(void *user, KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) {
+	if (RAD *rad = static_cast<RAD *>(user)) {
+		lilu.onKextLoadForce(&kextRadeonSupport);
+	}
+}
 
 RAD *RAD::callbackRAD;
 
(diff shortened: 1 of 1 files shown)


schreiberstein:
I can confirm this resolves the kernel panic on my system (Gigabyte Z790 Aero G, Tahoe Release, ASRock Radeon 6900 XT), and also did not cause any side effects on my main Sonoma system.

[WhateverGreen.kext.zip](https://github.com/user-attachments/files/22447013/WhateverGreen.kext.zip)

pyquick:
If you do this, it will be like having no patch at all. This will cause purple bars on the screen and loss of HIDPI, as if it were not patched.

royalgraphx:
> If you do this, it will be like having no patch at all. This will cause purple bars on the screen and loss of HIDPI, as if it were not patched.

What do you mean by this @pyquick ? You need to explain further, with logs, GPU, and what version of macOS you are talking about

pyquick:
This fixes the rec startup issue, but as I mentioned:

1. The screen briefly displays a purple bar after startup (only on AMD, I have a 590, but other systems should experience this as well).

2. There's no HIDPI at the login screen (at least the manual HIDPI setting doesn't work), only when you enter the desktop.


In summary: The above issue is not a problem with the official version (1.7.0); and it only occurs if WEG isn't loaded.

royalgraphx:
if WEG is not loaded... then it's not doing anything, so it wouldn't be resolved by 1.7.0 nor 1.7.1, you would need some logs near the time of that happening for me to really understand anything going on

WhateverGreen/kern_rad.cpp
```diff
@@ -80,6 +82,16 @@ static const char *powerGatingFlags[] {
 	"CAIL_DisableSAMUPowerGating"
 };
 
+/**
+ *  This function will be called by Lilu ONLY when IOGraphicsFamily has loaded.
+ *  Its only job is to then activate our patches for AMDSupport.
+ */
+static void onIOGraphicsLoad(void *user, KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) {
+	if (RAD *rad = static_cast<RAD *>(user)) {
+		lilu.onKextLoadForce(&kextRadeonSupport);
```
vit9696, line 91: `onKextLoadForce` APIs store kexts registered with this API in an internal structure, which is processed only once upon KernelPatcher registration. Calling `onKextLoadForce` here will be essentially a no-op and will never trigger `AMDSupport` and similar patches.

vit9696:
Thank you for your contribution! As I mentioned in the review comment, this change cannot work as is due to Lilu API design. Some other approach is necessary to solve Tahoe issues. I believe it is best to create a separate PR with a new solution rather than try to entirely change this one for history reasons. I will close this PR, but feel free to continue discussion here and/or file an issue in acidanthera/bugtracker for further discussion ^_^

royalgraphx:
No problem, i'll be sure to do some more investigation and understanding before submitting a separate PR, thanks for the review!

laobamac:
> Thank you for your contribution! As I mentioned in the review comment, this change cannot work as is due to Lilu API design. Some other approach is necessary to solve Tahoe issues. I believe it is best to create a separate PR with a new solution rather than try to entirely change this one for history reasons. I will close this PR, but feel free to continue discussion here and/or file an issue in acidanthera/bugtracker for further discussion ^_^

@vit9696 Hi, my friend. I found that when using BaseSystemKernelExtendes.kc as the kernel cache in the system, WhateverGreen immediately freezes at KextLog:... once it hooks AMDSupport. This affects the boot recovery mode and the second stage of OTA. But when starting the normal system, SystemKernelExtendes.kc is used, and WhateverGreen's patch for AMDSupport is working properly. What is the reason for this? Is there any difference between AMDSupport.kext in these two KCs?

Best regards.

royalgraphx:
When I diff'd them I saw zero differences, as far as I remember. The issue is still to be resolved, and frankly I haven't spent time looking into it, feel free to look into it and creating a new PR if required.

laobamac:
> When I diff'd them I saw zero differences, as far as I remember. The issue is still to be resolved, and frankly I haven't spent time looking into it, feel free to look into it and creating a new PR if required.

Yes, that's why I haven't found the real cause of this problem yet. However, delaying the loading of AMDSupport.text will cause WEG to render all its patches invalid.

We still need to know what factors prevent macOS from correctly accepting WEG hooks for AMDSupport when using BaseSystemKernelExtendes.kc as the kernel cache to start the system.

royalgraphx:
Yes, clearly lol

PR #126, opened 2025-10-29
Fix sticking at KernelCache validating on Tahoe

laobamac:
Fix "WEG Causes Boot Hang in macOS Tahoe Installer (Auqa Session issue), Works After Installation" https://github.com/acidanthera/bugtracker/issues/2509

### DIFF
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -10,6 +10,11 @@
 #include <Headers/kern_devinfo.hpp>
 #include <IOKit/IOService.h>
 
+#include <sys/vnode.h>
+#include <sys/mount.h>
+#include <sys/uio.h>
+#include <sys/namei.h>
+
 #include <Availability.h>
 #include <IOKit/IOPlatformExpert.h>
 
@@ -302,17 +307,53 @@ IOReturn RAD::wrapAMDRadeonX6000AmdRadeonFramebufferGetAttribute(IOService *fram
 	return ret;
 }
 
+bool fileExistsAtPath(const char *path) {
+	vnode_t vp = NULL;
+	vfs_context_t ctx = vfs_context_current();
+	int error = vnode_lookup(path, 0, &vp, ctx);
+	
+	if (error == 0 && vp != NULL) {
+		vnode_put(vp);
+		DBGLOG("rad", "File exists: %s", path);
+		return true;
+	}
+	
+	DBGLOG("rad", "File NOT found: %s (error: %d)", path, error);
+	return false;
+}
+
 bool RAD::isNormalSys() {
-	if (lilu.getRunMode() == 2) {
-		return false;
+	const char* modeStr = "Unknown";
+	switch (lilu.getRunMode()) {
+		case 1: modeStr = "Normal Boot"; break;
+		case 2: modeStr = "Recovery/Installer"; break;
+		case 4: modeStr = "Safe Mode"; break;
 	}
 
-	// OTAing will load "com.apple.recoveryosd"
-	OSDictionary *match = IOService::serviceMatching("com.apple.recoveryosd");
-	if (match) {
-		return false;
+	IOLog("WEG: isNormalSys() called | RunMode: %s (%d)\n", modeStr, lilu.getRunMode());
+
+	// Ignore all detections when forcibly enabling -radconnector
+	if (lilu.getRunMode() == 1 && checkKernelArgument("-radconnector")) {
+		IOLog("WEG: -radconnector detected in Recovery, FORCING override enable\n");
+		return true;
 	}
 
+	// Directly reject abnormal startup environments
+	if (lilu.getRunMode() != 1) {
+		IOLog("WEG: Not in Normal Boot mode, but RunMode=%d → blocking override\n", lilu.getRunMode());
+		return false;
+	}
+	
+	// When using BaseSystemKernelExtensions as the Kernel Cache, the linker is also not overwritten.
+	// Because in the second stage of the OTA (when the macOS Installer finishes booting and the volume
+	// label returns to Macintosh HD), it still uses BaseSystemKernelExtensions.kc, but Lilu mistakenly determines
+	// it as a normal boot. This causes it to still freeze during the KC verification stage.
+	if (fileExistsAtPath("/System/Library/KernelCollections/BaseSystemKernelExtensions.kc")) {
+			IOLog("WEG: BaseSystemKernelExtensions.kc EXISTS -> BLOCK override (prevent freeze)\n");
+			return false;
+		}
+
+	IOLog("WEG: Normal boot -> ALLOW connector override\n");
 	return true;
 }
 
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -306,8 +306,14 @@ bool RAD::isNormalSys() {
 	if (lilu.getRunMode() == 2) {
 		return false;
 	}
-	else
-		return true;
+
+	// OTAing will load "com.apple.recoveryosd"
+	OSDictionary *match = IOService::serviceMatching("com.apple.recoveryosd");
+	if (match) {
+		return false;
+	}
+
+	return true;
 }
 
 bool RAD::ifNeedOverrideConnector(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) {
--- a/WhateverGreen/kern_rad.cpp
+++ b/WhateverGreen/kern_rad.cpp
@@ -28,7 +28,6 @@ static const char *pathRadeonX4200[]        { "/System/Library/Extensions/AMDRad
 static const char *pathRadeonX4250[]        { "/System/Library/Extensions/AMDRadeonX4250.kext/Contents/MacOS/AMDRadeonX4250" };
 static const char *pathRadeonX5000[]        { "/System/Library/Extensions/AMDRadeonX5000.kext/Contents/MacOS/AMDRadeonX5000" };
 static const char *pathRadeonX6000[]        { "/System/Library/Extensions/AMDRadeonX6000.kext/Contents/MacOS/AMDRadeonX6000" };
-static const char *pathAMDRadeonServiceManager[] { "/System/Library/Extensions/AMDRadeonServiceManager.kext/Contents/MacOS/AMDRadeonServiceManager" };
 static const char *patchPolarisController[] { "/System/Library/Extensions/AMD9500Controller.kext/Contents/MacOS/AMD9500Controller" };
 
 static const char *idRadeonX3000New {"com.apple.kext.AMDRadeonX3000"};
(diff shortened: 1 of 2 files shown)


WhateverGreen/kern_rad.cpp
```diff
@@ -302,6 +307,21 @@ IOReturn RAD::wrapAMDRadeonX6000AmdRadeonFramebufferGetAttribute(IOService *fram
 	return ret;
 }
 
+bool RAD::ifNeedOverrideConnector(KernelPatcher &patcher, size_t index, mach_vm_address_t address, size_t size) {
+	// No need to overwrite the connector when booting BaseSystem on macOS Tahoe.
+	// This step is unnecessary and will cause macOS Tahoe recovery mode to freeze when booting.
+	// Committed by laobamac
+	if (getKernelVersion() >= KernelVersion::Tahoe) {
+		if (kextRadeonServiceManager.loadIndex != index) {
```
averycblack: ifNeedOverrideConnector is only ever called when index == kextRadeonSupport.loadIndex. Does this not have the affect of removing all connector patches in all Tahoe versions?

laobamac:
> ifNeedOverrideConnector is only ever called when index == kextRadeonSupport.loadIndex. Does this not have the affect of removing all connector patches in all Tahoe versions?

Yes, because RadeonServiceManager will load normally when macOS starts normally.

---------

We can change it to a simpler way, see the latest commit.

laobamac:
Now it works perfectly on macOS Tahoe, supporting installer/recovery mode/OTA.

laobamac:
It turns out that it works very stably on macOS Tahoe and has been tested a lot. The following is the feedback from a small number of users.
https://www.insanelymac.com/forum/topic/361713-pre-release-macos-tahoe-261/page/10/#findComment-2843066

https://www.insanelymac.com/forum/topic/361713-pre-release-macos-tahoe-261/page/10/#findComment-2843098

https://www.insanelymac.com/forum/topic/361713-pre-release-macos-tahoe-261/page/9/#findComment-2843040

https://www.insanelymac.com/forum/topic/361713-pre-release-macos-tahoe-261/page/9/#findComment-2843051

pyquick:
Yes, it could work well on Tahoe&rec

WhateverGreen/kern_rad.cpp
```diff
@@ -302,6 +307,73 @@ IOReturn RAD::wrapAMDRadeonX6000AmdRadeonFramebufferGetAttribute(IOService *fram
...
+	if (lilu.getRunMode() != 1) {
+		IOLog("WEG: Not in Normal Boot mode, but RunMode=%d → blocking override\n", lilu.getRunMode());
+		return false;
+	}
+	
+	// When using BaseSystemKernelExtensions as the Kernel Cache, the linker is also not overwritten.
+	// Because in the second stage of the OTA (when the macOS Installer finishes booting and the volume
+	// label returns to Macintosh HD), it still uses BaseSystemKernelExtensions.kc, but Lilu mistakenly determines
+	// it as a normal boot. This causes it to still freeze during the KC verification stage.
+	if (fileExistsAtPath("/System/Library/KernelCollections/BaseSystemKernelExtensions.kc")) {
```
vit9696, line 351: I believe we want this to be fixed on Lilu side.

laobamac: > I believe we want this to be fixed on Lilu side.

As of now, only this patch has encountered issues on macOS Tahoe. Perhaps modifying WhateverGreen is already simple enough, there is no need to modify Lilu temporarily

WhateverGreen/kern_rad.cpp
```diff
@@ -333,7 +405,13 @@ bool RAD::processKext(KernelPatcher &patcher, size_t index, mach_vm_address_t ad
 	}
 
 	if (kextRadeonSupport.loadIndex == index) {
-		processConnectorOverrides(patcher, address, size, true);
+		// The abnormal start-up system does not cover the connector on Tahoe.
+		if (ifNeedOverrideConnector(patcher, index, address, size) != false) {
```
vit9696, line 409: After we fix Lilu to correctly detect normal mode I believe the change here should look as follows:

```c
if (getKernelVersion() < KernelVersion::Tahoe
  || checkKernelArgument("-radconnector")
  || lilu.getRunMode() == LiluAPI::RunningNormal) {
  processConnectorOverrides(patcher, address, size, true);
  DBGLOG("rad", "processing override connector");
} else {
  DBGLOG("rad", "skipping override connector");
}
```

No other changes should be necessary.

laobamac: > After we fix Lilu to correctly detect normal mode I believe the change here should look as follows:
> 
> ```c
> if (getKernelVersion() < KernelVersion::Tahoe || checkKernelArgument("-radconnector") || lilu.getRunMode() == LiluAPI::RunningNormal) {
>   processConnectorOverrides(patcher, address, size, true);
>   DBGLOG("rad", "processing override connector");
> } else {
>   DBGLOG("rad", "skipping override connector");
> }
> ```
> 
> No other changes should be necessary.

I understand, do you mean to put the method that recognizes BaseSystemKernelExtendes.kc into Lilu's getRunMode?

vit9696: Well, we could go this way if no better is found. Please make sure we guard it with macOS 26, so that we definitely do not regress on older operating systems. Ideally we want a lighter approach.

laobamac: > Well, we could go this way if no better is found. Please make sure we guard it with macOS 26, so that we definitely do not regress on older operating systems. Ideally we want a lighter approach.

But when Lilu.kext checks the running status, the root file system may not have been loaded yet, and using vfs_lookup at this time will cause a panic.

vit9696: In this case I think we should find some other approach. Do you have an `ioreg` dump? Ideally both `IODeviceTree` and `IOService` planes.

laobamac: > In this case I think we should find some other approach. Do you have an `ioreg` dump? Ideally both `IODeviceTree` and `IOService` planes.

Unfortunately, ioreg only recorded BootKernelExtensions.kc. And there is no other obvious difference except BaseSystemKernelExtendes.kc.

vit9696: Hmmm, could you upload one for me to think about it as well? If it contains private data, you could also e-mail me to vit9696 at pm dot me ^^

laobamac: > Hmmm, could you upload one for me to think about it as well? If it contains private data, you could also e-mail me to vit9696 at pm dot me ^^

OK， Wait for me a moment, I will restart into Recovery and save one

laobamac:
[normal.zip](https://github.com/user-attachments/files/23291701/normal.zip)

[rec.txt.zip](https://github.com/user-attachments/files/23291700/rec.txt.zip)

@vit9696

laobamac:
The system in the second stage of OTA is unable to capture IOReg. But in fact, we just want to know how to implement the OTA phase😂

vit9696:
Hmm, that is problematic, SSH also does not work?

laobamac:
> Hmm, that is problematic, SSH also does not work?

Yes, the entire system crashed directly.
It's not ruled out that it's a problem with my machine, maybe you can also try OTA.

vit9696:
Also, your recovery dump does not include IODeviceTree plane, thus e.g. `chosen` is missing. The command should be `ioreg -l -p IODeviceTree`.

laobamac:
[rec2.txt.zip](https://github.com/user-attachments/files/23291929/rec2.txt.zip)

here

vit9696:
From this I see several candidates for OTA test:

* We could try to check whether `IODeviceTree:/chosen/booter-name` contains `bootbase.efi` or `boot.efi`.
* We could try to check for `IODeviceTree:/chosen/root-snapshot-name` presence.
* We could try to check whether `IODeviceTree:/chosen/boot-file` contains `boot\System\Library\KernelCollections\BootKernelExtensions.kc` or something else.

Ideally we need an IOReg dump from OTA v2 stage though, hard to be sure otherwise.

laobamac:
> From this I see several candidates for OTA test:
> 
> * We could try to check whether `IODeviceTree:/chosen/booter-name` contains `bootbase.efi` or `boot.efi`.
> * We could try to check for `IODeviceTree:/chosen/root-snapshot-name` presence.
> * We could try to check whether `IODeviceTree:/chosen/boot-file` contains `boot\System\Library\KernelCollections\BootKernelExtensions.kc`
or something else.
> 
> Ideally we need an IOReg dump from OTA v2 stage though, hard to be sure otherwise.

These are all unavailable, I have already tried them here. Because the second phase of OTA has already used boot.efi and BootSystemExtendes.kc, BaseSystemKernelExtendes.kc will only be loaded just before initializing the user interface.

This is my attempt from last month, and my thoughts are the same as yours. But it is not available.[kern_extra.zip](https://github.com/user-attachments/files/23292021/kern_extra.zip)

laobamac:
My friend, I will come back tomorrow to see this issue. It's already 2:10 in the morning in China, I'll go rest first.

bye

QiuChenly:
CleanShot 2026-01-13 at 13 31 58

---

**该补丁是不必要的。**

问题的根本原因在于 Lilu 框架，而非 WhateverGreen Kext。

具体而言，Lilu 在进行代码注入时发生了越界写入，错误地覆盖了 AMDSupport 的常规初始化代码。这一问题会导致在启动最新的 macOS Tahoe 版本（通过 OTA 更新）时触发内核崩溃（Kernel Panic）。

在旧版本的 macOS 上，AMDSupport 模块通常具有足够的代码空间容纳注入的代码，因此不会出现此问题。

我已测试了针对 Lilu 库的修复提交：
https://github.com/acidanthera/Lilu/pull/102

该修复可以正常工作，进一步证实了核心问题出在 Lilu 而非 WhateverGreen Kext。

---

**This patch is unnecessary.**

The root cause of the issue lies in the Lilu framework, not in WhateverGreen Kext.

Specifically, Lilu performs an out-of-bounds write during code injection, which incorrectly overwrites the regular initialization code of AMDSupport. This causes a Kernel Panic when booting the latest macOS Tahoe version (via OTA update).

On older versions of macOS, the AMDSupport module typically has sufficient code space to accommodate the injected code, so this issue does not occur.

I have tested the fix commit for the Lilu library:
https://github.com/acidanthera/Lilu/pull/102

This fix works correctly, further confirming that the core issue is with Lilu, not WhateverGreen Kext.

laobamac:
After passing the 26.3 (b) Beta (25D770870v) test, the recovery mode and normal system can be booted normally.
The speed of the boot recovery mode is obviously slower (compared to the case of directly disabling the overlay connector in the recovery mode). I'm not sure what the reason is, but it should be my own reason. After all, the principle alone will not affect the startup speed.

After https://github.com/acidanthera/Lilu/pull/102 is merged, I will close this PR.

Thanks to @averycblack and @vit9696 for everything they have done.

Eggmanplant:
> CleanShot 2026-01-13 at 13 31 58
> 
> **该补丁是不必要的。**
> 
> 问题的根本原因在于 Lilu 框架，而非 WhateverGreen Kext。
> 
> 具体而言，Lilu 在进行代码注入时发生了越界写入，错误地覆盖了 AMDSupport 的常规初始化代码。这一问题会导致在启动最新的 macOS Tahoe 版本（通过 OTA 更新）时触发内核崩溃（Kernel Panic）。
> 
> 在旧版本的 macOS 上，AMDSupport 模块通常具有足够的代码空间容纳注入的代码，因此不会出现此问题。
> 
> 我已测试了针对 Lilu 库的修复提交： [acidanthera/Lilu#102](https://github.com/acidanthera/Lilu/pull/102)
> 
> 该修复可以正常工作，进一步证实了核心问题出在 Lilu 而非 WhateverGreen Kext。
> 
> **This patch is unnecessary.**
> 
> The root cause of the issue lies in the Lilu framework, not in WhateverGreen Kext.
> 
> Specifically, Lilu performs an out-of-bounds write during code injection, which incorrectly overwrites the regular initialization code of AMDSupport. This causes a Kernel Panic when booting the latest macOS Tahoe version (via OTA update).
> 
> On older versions of macOS, the AMDSupport module typically has sufficient code space to accommodate the injected code, so this issue does not occur.
> 
> I have tested the fix commit for the Lilu library: [acidanthera/Lilu#102](https://github.com/acidanthera/Lilu/pull/102)
> 
> This fix works correctly, further confirming that the core issue is with Lilu, not WhateverGreen Kext.

is this gpu lexa or baffin

Eggmanplant:
it got mergeddddddddddd!!!!!! works!!!!

PR #127, opened 2026-08-04
Fix boot panic on macOS 26 (Tahoe) with AMD GPUs

olwimo:
## Problem

On macOS 26 (Tahoe), WhateverGreen panics the kernel while it is enumerating PCI and USB devices whenever an AMD GPU is present. The panic lands before the panic log can reach disk, so it leaves nothing behind.

Four separate defects turned out to be involved. Each only became reachable once the previous one was fixed, which is why this is a single PR rather than four.

## 1. Corrupt trampoline on `IORegistryEntry::getProperty`

`KernelPatcher::createTrampoline` copies displaced prologue bytes verbatim — the comment there reads `// Copy the prologue, assuming it is PIC`. An instruction carrying a relative operand inside that range therefore resolves against the trampoline once moved.

On Tahoe, `IORegistryEntry::getProperty(const char *)` has only 13 bytes of movable prologue:

```
0:  55           pushq %rbp
1:  4889e5       movq  %rsp, %rbp
4:  4156         pushq %r14
6:  53           pushq %rbx
7:  4889fb       movq  %rdi, %rbx
10: 4889f7       movq  %rsi, %rdi
13: e87edbfbff   callq OSSymbol::withCString    <-- rel32
```

Reaching the 14 bytes an absolute jump needs swallows that `callq`. #122 switched this route to `routeMultipleLong`, which forces an absolute route, so the trampoline's call to `OSSymbol::withCString` lands on unrelated memory and the kernel dies on the first property lookup.

The route now selects a jump type the prologue can actually survive: `routeMultipleLong` when a 14-byte displacement is clean, otherwise `routeMultipleShort` — which refuses to patch when the callback is out of reach instead of quietly widening the jump — and skipping the route entirely when neither is safe.

Every route WhateverGreen makes was audited the same way against the real Tahoe binaries, using Lilu's own hde64 decoder. `getProperty` is the only hazard; `setProperty` (exactly 14 movable bytes) and `cs_validate_page` (16) are unaffected.

## 2. Re-entrancy in the `getProperty` wrapper

A valid trampoline was still fatal. `wrapGetProperty` calls `getParentEntry` on every dictionary-valued lookup, re-entering the registry while a walk of that same registry is in progress — including the walk Lilu performs to build `DeviceInfo`. On a machine with a long PCI device list it dies partway through.

The wrapper can only ever do useful work when `CFG,` / `PP,` / `CAIL,` overrides were injected, so it is no longer routed unless one is actually present. `-radnoprop` disables both property routes outright.

## 3. Calls through unresolved trampolines

Making `getProperty` conditional left `orgGetProperty` null, and `wrapSetProperty` called straight through it whenever a `model` property was set — which `AMDRadeonX6000` does when it loads.

`updateConnectorsInfo` has the same defect via `orgGetAtomObjectTableForType`. That symbol, along with `AtiAtomBiosDceInterface::getConnectorsInfo`, `AMDLegacyController::start`, `AppleGraphicsDevicePolicy::start`, `_dce_driver_set_backlight` and `_dce_panel_cntl_hw_init`, **does not exist on macOS 26 at all**. Both call sites are now checked.

## 4. Unbounded framebuffer write

`wrapFramebufferInit` sized its back-copy and zero-fill from the console `vinfo` but wrote into the framebuffer's VRAM mapping:

```c
memset(dst, 0, info.v_rowbytes * info.v_height);
```

The guard above it compares against the display mode's pixel information, which is not the same thing as the mapping. The write is now clamped to `IOMemoryMap::getLength()`.

## Testing

Built for Release and Debug. Verified on a Radeon Pro W5700 (Navi 10, `1002:7312`) running macOS 26.6 (25G72), booting from an unbootable starting state through to a full desktop on stock settings (`agdpmod=pikera unfairgva=1`), with each fix bisected by boot argument along the way.

Only tested on that one machine and on Navi. The Intel and NVIDIA submodules were not exercised, though nothing here is specific to AMD except the gating in point 2.

## Note for maintainers

This includes the `MODULE_VERSION` bump to 1.7.2 and its Changelog entries, since the Changelog text refers to the version. Happy to drop both commits' version handling if you would rather bump separately, as the history suggests you normally do.


TonyC5:
Does this issue need to be added to Acidanthera's bugtracker [here](https://github.com/acidanthera/bugtracker/issues)?

PsychoTea:
Hi guys, I would just like to confirm this PR succeessfully helped me get up and running on Tahoe 26.6.1 (25G76) with an RX5700XT, using `agdpmod=pikera` after migrating from Sonoma.
Before using this build I was getting a mixture of panics on boot, hangs, and other failures around graphics initialization time. After installing this build, the card immediately started working fully working (all 4 outputs) with proper graphics acceleration.

Thanks for the PR @olwimo!

vit9696:
The correct way to use WEG on macOS 26 is `agdpmod=ignore`, from what I remember at least, this is how it works for me with RX 6600 and current WhateverGreen from master.

Fixing other modes on macOS 26 is unlikely necessary in the first place, but I might consider accepting some of the changes if they make sense. Claude hallucinations presented in this patch make little sense, because they give no proof the assumption made is actually correct.

Please start with posting a panic log with keepsyms=1. Let's iterate from there.

olwimo:
@vit9696 yes, this pr is Claude-ed, and certainly shouldn't be merged without proper (human) evaluation. The thing is, I have a cand.scient. in computer science, but I also have a heavier workload than ever, because everything is moving so fast with ai assistance, and two sons who need me as a parent and they tell me they can't make do with a robot-dad yet - so I don't have time to follow through on this hobby project, no matter how much I'd like to.
I *can* tell you the background for this pr, though:
I had panics on my intel 10900xe Hackintosh after sleep, because my amd gpu (Navi 10/Radeon Pro W5700) didn't like to wake up, my Tahoe build couldn't boot with weg, and no combination of bios settings that made any sense to me at face value worked (I haven't dived *that* deep into the acpi, so I might have missed an opportunity for an easy aml patch).
So, I set up a local weg cloned repo and asked Claude to fix it, basically. I had to reboot a few times, but mostly it ran on its own, and in the end it solved my sleep issues. I mostly made the pr so noone else would have to spend credits on the same, and I have enough professional experience to know that this is not fit for merge yet, not by a long shot. I was just handing over what I had, and pretty soon I have to get off the bus at my stop, and I won't have time for writing any more hobby-stuff until after the kids are asleep. When they are, then I'll gladly do a boot without weg, record the panic, and upload the log here.
Thanks for your work - when I retire and the kids move on to bigger and better things, then I'd love to spend more time with this, or a similar project, if humanity is still relevant then...

olwimo:
So...:
With stock WEG 1.7.0-Release from the acidanthera repo and agdpmod=ignore in boot-args on an otherwise working config, it hangs while enumerating PCI and USB devices (see attached photo, if you need me to run the debug versions of OC and relevant kexts and get it logged to disk, I'll do it, but I'm hoping it won't be necessary - it hangs there).
PXL_20260817_235440993

With WEG disabled and agdpmod boot-arg removed it boots fine - but after sleep it gives this kp:
[kp.txt](https://github.com/user-attachments/files/31160235/kp.txt)

With WEG built from this PR and boot-args agdpmod=pikera and unfairgva=1 it boots, it sleeps and it wakes.

I hope that's useful in some way.

vit9696:
Sorry for a slow reply, but it took me some time to get here. Let me go step by step on this PR:

> 1. Corrupt trampoline on IORegistryEntry::getProperty

* It is correct that on macOS 26 getProperty has only 13 bytes of movable prologue. At least I can confirm that with 26.6.2 kernel.
* It is also correct that #122 switched getProperty routing to `routeMultipleLong`, because only `routeMultipleLong` allows chaining routes. This change is in fact absolutely necessary for [iBridged](https://github.com/Carnations-Botanica/iBridged) to work. 
* It is not correct that `routeMultipleLong`, however, always needs 14 bytes for patches. Starting with Lilu 1.4.6 (https://github.com/acidanthera/Lilu/commit/bba464272a3599f064b11ef197b4d769ce00df92) `routeMultipleLong` transparently uses "slotted" medium jumps whenever possible, which are only 6 bytes long instead of large 14 byte long jumps.

To sum it up, for this problem on the following is true:

* The claim IORegistryEntry::getProperty trampoline is corrupt is entirely false and is just a Claude hallucination. You can safely return `routeMultipleLong`.
* _You are missing a Lilu update_. Slotted medium changes were broken at least in some macOS 26.x updates prior to Lilu 1.7.2 (https://github.com/acidanthera/Lilu/pull/102).
* Slotted medium jumps do not work for you for some reason. Easiest to check this is to dump a debug log via `liludump=120` or so.

> 2. Re-entrancy in the getProperty wrapper

The way the problem is described makes no sense to me:

* First it says there is a fatal reentrancy problem.
* Second it says it calls `getParentEntry` on every dictionary-valued lookup, even during various tree builds. This is indeed true, but I have no idea why it is relevant to the problem.
* Third it says it is fatal for on a machine with a long PCI device list without explaining why.

Nevertheless the garbage, the suggested change ignores the fact that CFG, / PP, / CAIL, _overrides may not be visible at a time routing happens_. This exact reason is why getProperty patch is unconditional, and this is why the suggested change is wrong.

`getParentEntry` is really fast, it is literally just a `getObject` call. But if anything, I guess you can try to speedup the function by checking the argument first. Like this:

<details>

```c
OSObject *RAD::wrapGetProperty(IORegistryEntry *that, const char *aKey) {
	auto obj = FunctionCast(wrapGetProperty, callbackRAD->orgGetProperty)(that, aKey);
	auto props = OSDynamicCast(OSDictionary, obj);

	if (props && aKey) {
		const char *prefix {nullptr};
		if (aKey[0] == 'a') {
			if (!strcmp(aKey, "aty_config"))
				prefix = "CFG,";
			else if (!strcmp(aKey, "aty_properties"))
				prefix = "PP,";
		} else if (aKey[0] == 'c' && !strcmp(aKey, "cail_properties")) {
			prefix = "CAIL,";
		}

		if (prefix) {
			auto provider = OSDynamicCast(IOService, that->getParentEntry(gIOServicePlane));
			if (provider) {
				DBGLOG("rad", "GetProperty discovered property merge request for %s", aKey);
				auto rawProps = props->copyCollection();
				if (rawProps) {
					auto newProps = OSDynamicCast(OSDictionary, rawProps);
					if (newProps) {
						callbackRAD->mergeProperties(newProps, prefix, provider);
						that->setProperty(aKey, newProps);
						obj = newProps;
					}
					rawProps->release();
				}

			}
		}
	}

	return obj;
}
```

</details>

> 3. Calls through unresolved trampolines

* `orgGetProperty` will not be NULL if you do not break it. This problem is introduced by the changes in your patch, and then you try to fix it.
* `orgGetAtomObjectTableForType` is not used at all on macOS High Sierra and newer. See `getKernelVersion() >= KernelVersion::HighSierra`.
* `wrapGetConnectorsInfoV2` is the only one possibly called on macOS 26. It explicitly passes `nullptr` to both `updateConnectorsInfo` parameters: `atomutils` and `gettable`, which guards any null pointer dereference.

I do not think any of that makes sense besides the fact that Claude broke `orgGetProperty` during prototyping, and then tried to excuse itself.

> 4. Unbounded framebuffer write

This change looks suspicious. I suspect this in fact could be the main reason it broke. However, more understanding is needed before any change is made.

Basically your patch says "let's not write more than is allocated". This is ok, but we write to a framebuffer, which must be allocated there. If it is not allocated there, well, we are doomed and have a memory corruption before that code even runs.

Could you maybe trigger a panic when mapped memory is less than framebuffer memory and perhaps print:

* mapped memory
* v_rowbytes
* v_height

At least this way we can compare numbers. The original code is a bit weird, and I need to remember why it is written this way in the first place.

------------

TL;DR

* Unsure whether you can feed the above to Claude and get sensible results, but you could try at least.
* I definitely want you [try to] revert changes 1, 2, and 3. Most likely it will work without them if you make sure to update Lilu. You can try the optimisation I suggested for change 2. It is not strictly necessary but could result in boot time improvement.
* Change 4 is wrong per se, but if the problem stated indeed exists, some fix is necessary. I suggested first steps to explore the problem.

olwimo:
I'll get to it asap, I haven't done systems programming at all since school, and that was only on a minix-like for mips(I think?) arch, so... not really anything relevant. Safe to say I'll still have to rely quite a bit on claude, but I will spend some real time looking at the code it writes and questioning the reasoning this time. Anyway, just one correction: I *do* have Lilu 1.7.2 installed and running:

% grep -A1 'CFBundleShortVersionString' /Volumes/EFI/EFI/OC/Kexts/Lilu.kext/Contents/Info.plist
	<key>CFBundleShortVersionString</key>
	<string>1.7.2</string>

% diff Lilu.kext/Contents/MacOS/Lilu /Volumes/EFI/EFI/OC/Kexts/Lilu.kext/Contents/MacOS/Lilu 

(Downloaded 1.7.2 fresh off the github release page, diff returned empty)

... and the decisive test:
% kextstat| grep 'Lilu'
Executing: /usr/bin/kmutil showloaded
No variant specified, falling back to release
   56   13 0xffffff8004c78000 0x31000    0x31000    as.vit9696.Lilu (1.7.2) C2A303B8-6D22-3163-A905-5DB8E340D285 <9 7 6 3 2 1>
   
... so, yeah. But I'll try to revert/rewrite the changes and test them, as well as I can and according to your comment, in a few moments.

olwimo:
Ok, tl;dr: I'm an idiot, these patches do nothing for me, I'd close this at once if it wasn't for @PsychoTea 's report.

Slightly longer (and embarrassing) story: Claude immediately ceded that reverting on 1 and 3 should do nothing; and it didn't. Neither did reverting most of 2 - or all of it and reverting 4 as well, running a functional copy of master with just a few diagnostics helpers added. At this point Claude noticed that I had upgraded from 26.6.1 to 26.6.2 since I last observed the boot failure, and was trying to talk me into re-installing that version to investigate further, and I was quite frankly not happy about that proposition. But then it clicked: oh, *yeah*, but *did* I actually run the weg master when I couldn't boot? 'Cause it was from a different repo, the acidanthera, not my recent fork. And yeah, cd to that repo and:

% git status
HEAD detached at 1.7.0
nothing to commit, working tree clean

... I'm sorry for wasting your time. I blame senile dementia, and will seek out a doctor for a proper treatment asap - or maybe just ask Claude for his professional opinion as a medical advisor.

@PsychoTea are you *sure* these patches work better than master for you?... because if they do, that's great, but completely accidental.

PsychoTea:
Hi @olwimo 

Put simply, it wasn’t booting with 1.7.0 from the releases, and worked after building and installing this PR. It’s very possible there was some other small change I made when installing this PR (ie. changing a boot arg), meaning my test may not have fully isolated this build. 

Taking a look at the commit history there is also ed5e710a5a1f15369e4284865d8510525fa3c357 which was not made available in releases. It’s possible this change alone fixed whatever issue I was having. 

If it would be beneficial, I can go back and test 1.7.0, latest master, and this PR to try and isolate things.

